Issue #7, opened 2018-10-25
Display network errors to the users


zudov:
Currently, when ajax calls fail we get nice red messages in the console, but nothing indicates it to the user.

f-f:
I think also this one was fixed by #29, I'll close as well

PR #12, opened 2018-10-26
Timeout long running actions, so user doesn't get stuck with spinner

### LINKED ISSUE #6: Add error timeout on loading spinner
Currently if some error happens during the async execution tied to the loading spinner, it just stays there forever.

This is not nice, and we should add a timeout on the spinner, so if it has been on for e.g. more than 30 secs we stop it and display an error.

zudov:
This does the timeouting and prevents the spinner for staying indefinitely, but doesn't show anything to the user. Doing that will be a part of #7

Fixes #6

### DIFF
--- a/apps/mitt-konto/src/MittKonto/Main.purs
+++ b/apps/mitt-konto/src/MittKonto/Main.purs
@@ -3,20 +3,17 @@ module MittKonto.Main where
 import Prelude
 
 import Data.Array ((:))
-import Data.DateTime.Instant as Instant
 import Data.Either (Either(..))
 import Data.Foldable (foldMap)
 import Data.JSDate (JSDate, parse)
 import Data.Maybe (Maybe(..))
 import Data.String (toUpper)
-import Data.Time.Duration as Time
 import Effect (Effect)
 import Effect.Aff (Aff)
 import Effect.Aff as Aff
 import Effect.Class (liftEffect)
 import Effect.Class.Console (log)
 import Effect.Exception (error)
-import Effect.Now as Effect
 import Effect.Unsafe (unsafePerformEffect)
 import KSF.Footer.Component as Footer
 import KSF.Login.Component as Login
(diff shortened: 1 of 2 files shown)


apps/mitt-konto/src/MittKonto/Main.purs
```diff
@@ -87,27 +89,41 @@ loadingIndicator Loading =
...
+       Aff.delay $ Aff.Milliseconds 100.0
+       setLoading $ Just Loading
+     actionFiber <- Aff.forkAff do
+       -- In the meanwhile we run the action
+       result <- Aff.try action
+       -- then we give result to the handler (swallowing errors thrown by it)
+       Aff.apathize $ liftEffect $ handler result
+     timeoutFiber <- Aff.forkAff do
+       -- in yet another thread we sleep we are counting down till timeout
+       Aff.delay $ Aff.Milliseconds (30.0 * 1000.0)
```
f-f: Let's have the timeout in a `timeoutSeconds` binding in a `let` on top of the function

apps/mitt-konto/src/MittKonto/Main.purs
```diff
@@ -87,27 +89,41 @@ loadingIndicator Loading =
...
+     loadingFiber <- Aff.forkAff do
+       -- In this thread we wait a bit and then switch the spinner on
+       -- This prevents flickering if the action completes instantly
+       Aff.delay $ Aff.Milliseconds 100.0
+       setLoading $ Just Loading
+     actionFiber <- Aff.forkAff do
+       -- In the meanwhile we run the action
+       result <- Aff.try action
+       -- then we give result to the handler (swallowing errors thrown by it)
+       Aff.apathize $ liftEffect $ handler result
```
f-f: Why swallow errors? Let's at least log them (I'd love to send them through analytics or something else so we can collect them remotely), otherwise debugging async stuff will be horrible

apps/mitt-konto/src/MittKonto/Main.purs
```diff
@@ -87,27 +89,41 @@ loadingIndicator Loading =
...
+     SetState
+  -> (Either Error a -> Effect Unit) -- ^ result handler
+  -> Aff a -- ^ asynchronous action
+  -> Effect Unit
+runAffLoading setState handler action =
+   Aff.launchAff_ do
+     loadingFiber <- Aff.forkAff do
+       -- In this thread we wait a bit and then switch the spinner on
+       -- This prevents flickering if the action completes instantly
+       Aff.delay $ Aff.Milliseconds 100.0
```
f-f: Let's make this `delay` parametric, as it is currently.

The reason is that in some cases (e.g. `logout`) we do want to start the loading right away, because we know it's going to take some time anyways. In that specific case we were getting flickering with this `100.0` because we'd unset the state, so the view would start to change, and _then_ we'd get the spinner for a brief moment before ending the action (so the comment above this is only partially correct, and maybe we should document this happening there).
So we ended up making it `0.0`, and everything was nice. And in general it's nice if the caller has the possibility to tweak this.

apps/mitt-konto/src/MittKonto/Main.purs
```diff
@@ -87,27 +89,41 @@ loadingIndicator Loading =
...
-         (\loading -> aff)
+-- | Allows to run the asynchronous action while showing the loading indicator
+--   and handling the result.
+runAffLoading
+  :: forall a.
+     SetState
+  -> (Either Error a -> Effect Unit) -- ^ result handler
+  -> Aff a -- ^ asynchronous action
+  -> Effect Unit
+runAffLoading setState handler action =
```
f-f: This `Loading` in the name is a bit unclear, let's make this `runAffWithSpinner`

apps/mitt-konto/src/MittKonto/Main.purs
zudov: Woah, thanks for explaining the origin for `0.0` delay, it now makes sense :D 

> we'd unset the state, so the view would start to change, and then we'd get the spinner

I think this is a bug in the first place. The user logged-in state shouldn't be unset until the logout procedure is actually finished. Or rather it should be done as the last thing in the async call (so that when the loading spinner disappears we are already back to the login screen). 

I'm actually currently digging in the logout related code, so I think I see why this came up. There's some affing there that doesn't seem correct, particularly in Facebook's SDK wrapping

Setting the delay to 0.0 just allows to hide the early state change behind the spinner. In case when there is no need to do any blocking async stuff (e.g. no facebook or google login, all that's needed to be done is to wipe the local storage) the user would see a flickering loading spinner if the delay is 0.0

> then we'd get the spinner for a brief moment before ending the action

This is a very nice catch, I haven't considered at all that if the action takes e.g. 150 milliseconds and the spinner is shown after 100 milliseconds it would actually flicker back after 50 milliseconds. So we should make it so that the spinner is only shown after delay, but once it's shown it stays for some half a second (even if the async action finishes before that). I guess ultimately this problem could be solved by clever spinner design (e.g. some kind of fade in/out animation or something), but putting it into some sensible time bounds should be useful in any case.

> And in general it's nice if the caller has the possibility to tweak this.

The "caller" here would be the child component that calls `props.launchAff_`, so far we were specifying that delay from the parent component when we were constructing the `launchAff_` that would be given to a child component. If we were to give them the possibility to tweak this, I think it should be in the form of providing `launchAff_` with the bounds of "how long will the action take" (e.g. `{ min: 100.0, max: 10.0 * 1000 }`). However, I don't think there's a need for that since user's tolerance to flickering and actions that hung and take forever is more-or-less global. More importantly, it might be quite confusing to specify the delays whenever we make async calls (since we won't really remember what do those numbers mean and how they affect the thing), but we might end up with a lot of `0.0` in places where it seems to fix problems.

f-f: > I think this is a bug in the first place. The user logged-in state shouldn't be unset until the logout procedure is actually finished. Or rather it should be done as the last thing in the async call (so that when the loading spinner disappears we are already back to the login screen).
> I'm actually currently digging in the logout related code, so I think I see why this came up. There's some affing there that doesn't seem correct, particularly in Facebook's SDK wrapping

Good catch, the reason why we made it this was that we wanted to unset the user first, before logging out FB and Google, as sometimes there would be hiccups. Or something like that 😅

But if you're digging into that logout code then yeah, let's apply all your suggestions 👍

PR #20, opened 2018-11-06
Paralellize and sanitize the logout. Make it JS-ready.

zudov:


f-f:
@zudov I merged because it's good anyways, but this doesn't fix the logout getting stuck and not removing the user's profile (I am now suspecting that SSO starts to log the user in again after we empty the state...)

(Browser: Firefox)

Issue #22, opened 2018-11-08
SSO and third party cookies


zudov:
Apparently SSO doesn't work if the browser that don't allow third-party access to cookies and site data. When "Accept third-party cookies and site data" option is set to never, the Janrain's side-loaded script fails with `SecurityError` when trying to do `getItem` (local storage):

![image](https://user-images.githubusercontent.com/2613966/48191936-19d34d80-e34f-11e8-82ec-f6eb5fc3babb.png)

The same happens e.g. on hbl.fi and prenumerera.

In mitt-konto it becomes a reason for a nasty bug. When logging out, mitt-konto runs `janrain_sso.end_session(callback)`, and shows the spinner. When `end_session` calls `callback`, mitt-konto removes the spinner and removes the user from the state.

The trouble is, that when `end_session` is broken by `SecurityError`s our `callback` is never called, so the spinner spins until a timeout (30s) and the user's state is never wiped. Reloading the page wipes the state and you end up correctly on the login page.

I can think of a couple of ways of handling this correctly, I'll try it out.

PR #25, opened 2018-11-09
Only do JanrainSSO.end_session() when SSO have actually worked

zudov:
A fix for #22 which amounts to following:

- When SSO succeeds to capture (`capture_success`) and/or update session (`callback_success`) we set `localStorage.KSF_JANRAIN_SSO_SUCCESS`
- We don't call `end_session` unless `KSF_JANRAIN_SSO_SUCCESS` is set (as there is no need)

### DIFF
--- a/packages/ksf-login/src/JanrainSSO.purs
+++ b/packages/ksf-login/src/JanrainSSO.purs
@@ -3,7 +3,7 @@ module JanrainSSO where
 import Prelude
 
 import Data.Either (Either(..))
-import Data.Maybe (Maybe(..))
+import Data.Maybe (Maybe(..), isJust)
 import Data.Nullable (Nullable)
 import Data.Nullable as Nullable
 import Effect (Effect)
@@ -12,8 +12,15 @@ import Effect.Aff as Aff
 import Effect.Class.Console as Console
 import Effect.Exception as Exception
 import Effect.Uncurried (EffectFn1, runEffectFn1)
+import LocalStorage as LocalStorage
 
 foreign import loadConfig :: Effect Config
+
+setSsoSuccess :: Effect Unit
+setSsoSuccess = LocalStorage.setItem "KSF_JANRAIN_SSO_SUCCESS" "true"
+
+getSsoSuccess :: Effect Boolean
+getSsoSuccess = isJust <$> LocalStorage.getItem "KSF_JANRAIN_SSO_SUCCESS"
 
 foreign import sso ::
   { check_session
--- a/packages/ksf-login/src/JanrainSSO.purs
+++ b/packages/ksf-login/src/JanrainSSO.purs
@@ -35,7 +35,7 @@ endSession :: Aff Unit
 endSession = Aff.makeAff \callback -> do
   Exception.catchException
     (\err -> do
-      Console.error $ "JanrainSSO.set_session failed: " <> Exception.message err
+      Console.error $ "JanrainSSO.end_session failed: " <> Exception.message err
       callback $ Left err)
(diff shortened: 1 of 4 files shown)


packages/ksf-login/src/JanrainSSO.purs
```diff
@@ -12,9 +12,19 @@ import Effect.Aff as Aff
 import Effect.Class.Console as Console
 import Effect.Exception as Exception
 import Effect.Uncurried (EffectFn1, runEffectFn1)
+import LocalStorage as LocalStorage
 
 foreign import loadConfig :: Effect Config
 
+setSsoSuccess :: Effect Unit
+setSsoSuccess = LocalStorage.setItem "KSF_JANRAIN_SSO_SUCCESS" "true"
```
f-f: Let's add logging here in all these 3 functions, so we know if/when they happen

packages/ksf-login/src/KSF/Login/Component.purs
```diff
@@ -129,10 +129,12 @@ receiveProps { props, state, setState, isFirstMount } = when isFirstMount do
              { callback_failure: mkEffectFn1 \a -> do
                 Console.log "Janrain SSO failure"
              , callback_success: mkEffectFn1 \a -> do
+                JanrainSSO.setSsoSuccess
                 Console.log "Janrain SSO success"
              , capture_error: mkEffectFn1 \a -> do
                 Console.log "Janrain SSO capture error"
              , capture_success: mkEffectFn1 \r@({ result: { accessToken, userData: { uuid } } }) -> do
+                JanrainSSO.setSsoSuccess
```
f-f: Do we need to set it in both of them?

zudov:
> This doesn't fix it for me :(

Yes, I've also noticed that it's not 100% reliable and let it rest for the weekend.

packages/ksf-login/src/KSF/Login/Component.purs
zudov: `capture_success` is called when we call `check_session` and it captures the user's session left by the other site. `callback_success` is called when we call `set_session` (yes, it seems like `check_session` saves that callback somewhere so it can be called by `set_session`), weird, but it seems to work.

zudov: The way I understand it works:

- We call `SSO.check_session { callback_success, capture_success }`
- If SSO fires, `capture_success` is executed with all needed info
- otherwise, the user provides credentials to Persona and it gives back the SSO code
- We call `SSO.set_session(ssoCode) which executes `callback_success`

This seemed as a strange flow at first, but I think I now understand the reasoning behind it. You can register all your SSO-related callbacks in one place (when calling `check_session`) which is very useful if they need to share some context. And then drive the thing with non-blocking calls like `set_session` and `end_session` from other places.

zudov:
> I left some comments so we add logging and eventually pin it down

Right. I wrote my understanding of how it works [up there](I left some comments so we add logging and eventually pin it down) , but I only know it from the documentation and observing mitt-konto. Let's confirm it with detailed logging.

zudov:
After much debugging and confusion since this works well locally, but not on netlify preview, we found that the same fix is needed for facebook SDK. When running on not-whitelisted (in Facebook app settings) domain calling `facebookLogout` would load the iframe, log the warning about not being whitelisted and never call our callback or notify about the error:

2018-11-15-125940_747x50_scrot

This would need the same fix as we did here (write to localStorage when facebook login succeeds, don't call facebook logout if it didn't). We'll do it in another PR.

PR #36, opened 2018-11-15
Do logoutFacebook only if we've had a facebook login

### LINKED ISSUE #22: SSO and third party cookies

zudov:
See #25 for context. This is pretty much the same, but for facebook.

Fix #22

### DIFF
--- a/packages/ksf-login/src/KSF/Login/Component.purs
+++ b/packages/ksf-login/src/KSF/Login/Component.purs
@@ -21,6 +21,7 @@ import Effect.Exception (Error)
 import Effect.Uncurried (EffectFn1, mkEffectFn1, runEffectFn1)
 import JanrainSSO as JanrainSSO
 import KSF.Login.Facebook.Sdk as FB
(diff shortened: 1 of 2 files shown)


zudov:
:tada:

PR #45, opened 2018-12-04
Check for FB object before loading Facebook SDK

### LINKED ISSUE #44: Check for FB object before loading Facebook script
Let's not load the FB script in `ksf-login` component if it's already been loaded some other way.

vapaj:
If the Facebook SDK is already loaded to the site, loading and initializing it the second time will cause the SDK to break.

Fixes #44

### DIFF
--- a/packages/ksf-login/src/KSF/Login/Facebook/Sdk.js
+++ b/packages/ksf-login/src/KSF/Login/Facebook/Sdk.js
@@ -54,15 +54,17 @@ exports._init = function(callback) {
           window.isFBLoaded = true;
           callback(FB)(); // callback itself returns an effect
         };
-        (function(d, s, id){
-          var js, fjs = d.getElementsByTagName(s)[0];
-          if (d.getElementById(id)) {return;}
-          js = d.createElement(s);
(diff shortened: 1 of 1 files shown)


packages/ksf-login/src/KSF/Login/Facebook/Sdk.js
```diff
@@ -45,7 +45,7 @@ exports._api = function(fb, path, method, params, callback) {
 exports._init = function(callback) {
   return function(fbConfig) {
     return function() { // returns an effect
-      if (window.isFBLoaded) {
+      if (window.isFBLoaded || typeof FB === 'object') {
```
zudov: This will evaluate to `true` if `FB` is `null` (`typeof null === "object"`). It probably doesn't matter since `FB` would be either `undefined` (when not loaded) or an actual object (when loaded), but it also might become a very gnarly bug.

Let's at least check for presence of `FB` (`FB && typeof FB === 'object'`). Maybe `FB` has some field that we can check for to find out its status?

Also, how about doing `window.FB` everywhere so that it's visible that its global?

Issue #40, opened 2018-11-23
Link to cancellation form


kirstipa:
Could we add a link "Avsluta din prenumeration" to Mitt konto? It should be right under the subscriptions.

Link: https://ksfmedia1.typeform.com/to/zbh3kU

(Icon to use is svg file and not supported here. I'll send it to you when needed.)

f-f:
Fixed in #46

PR #46, opened 2018-12-04
Add link to subscription cancellation form

vapaj:
Adds #40


kirstipa:
Looks good! I only now realised that maybe we should show the cancellation link only if user has a subscription. Let's do that before deploying to production.

PR #50, opened 2018-12-18
Subscription ui adjustments

### LINKED ISSUE #9: Visual adjustments for subscriptions
From @kirstipa

We should:
- [ ] have the list of subscriptions ordered so that the cancelled subscriptions will show at the bottom of the list
- [ ] remove the "nästa faktureringsdatum" from the cancelled subscriptions
- [ ] visually separate the active and inactive subscriptions (waiting for a layout for this)

Later we could consider showing the subscription period or the last day of subscription of those cancelled subscription.

vapaj:
Fix #9

### DIFF
--- a/apps/mitt-konto/src/MittKonto/Main.purs
+++ b/apps/mitt-konto/src/MittKonto/Main.purs
@@ -2,12 +2,13 @@ module MittKonto.Main where
 
 import Prelude
 
-import Data.Array (snoc, (:))
+import Data.Array (cons, foldl, snoc, zip, (:))
 import Data.Either (Either(..), either)
 import Data.Foldable (foldMap, oneOf)
 import Data.JSDate (JSDate, parse)
 import Data.Maybe (Maybe(..))
 import Data.String (toUpper)
+import Data.Tuple (Tuple(..))
 import Effect (Effect)
 import Effect.Aff (Aff)
 import Effect.Aff as Aff
@@ -163,6 +164,12 @@ userView { user } = React.fragment
   , classy DOM.div "col col-12 md-col-6 lg-col-6" [ subscriptionsView ]
   ]
   where
+    subsSorted = foldl canceledLast [] user.subs
+
+    canceledLast subs s
+      | Persona.isSubscriptionCanceled s = subs `snoc` s
+      | otherwise = s `cons` subs
+
     componentHeader title =
       classy DOM.span "mitt-konto--component-heading" [ DOM.text $ toUpper title ]
 
@@ -181,9 +188,11 @@ userView { user } = React.fragment
       componentBlock "Mina prenumerationer:" $ subscriptions <> [ break, subscribeImage ]
       where
         subscriptions =
-          case user.subs of
+          case subsSorted of
             [] -> [ componentBlockContent noSubscriptionsText ]
-            subs -> map (componentBlockContent <<< subscriptionView) subs `snoc` cancelSubscription
+            subs -> do
+              let subscriptionViews = subscriptionView <$> subs
+              (subscriptionComponentBlockContent <$> zip subs subscriptionViews) `snoc` cancelSubscription
 
     subscriptionView subscription =
       Subscription.subscription { subscription }
@@ -232,6 +241,14 @@ userView { user } = React.fragment
             componentHeader headerText
             : content
         }
+
+    subscriptionComponentBlockContent (Tuple subscription subView)
+      | Persona.isSubscriptionCanceled subscription =
+          DOM.div
+            { className: "mitt-konto--canceled-subscription"
+            , children: [ componentBlockContent subView ]
+            }
+      | otherwise = componentBlockContent subView
 
     componentBlockContent child =
        DOM.div
--- a/apps/mitt-konto/src/MittKonto/Main.purs
+++ b/apps/mitt-konto/src/MittKonto/Main.purs
@@ -239,6 +239,8 @@ userView { user } = React.fragment
             : content
         }
 
+    -- | Wraps canceled subscriptions.
+    --   Meaning, if the subscription has a canceled state, we want to add extra css to it.
     subscriptionComponentBlockContent (Tuple subscription subView)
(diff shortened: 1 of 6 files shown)


apps/mitt-konto/src/MittKonto/Main.purs
```diff
@@ -163,6 +164,12 @@ userView { user } = React.fragment
   , classy DOM.div "col col-12 md-col-6 lg-col-6" [ subscriptionsView ]
   ]
   where
+    subsSorted = foldl canceledLast [] user.subs
+
+    canceledLast subs s
```
vapaj: Note that this reverses the order of active subscriptions. If the subscriptions are in no particular order coming from Persona, this works fine, otherwise, this should be done in some other manner.

f-f: I wonder if we should just define an `Ord` for `SubscriptionState` so we can `sort`?

apps/prenumerera/src/Prenumerera/PaymentSelect.purs
```diff
@@ -93,10 +93,10 @@ didMount self@{ props: { location: { state: Just state } } } = do
       maybe (pure unit) (send self) maybeAction
 didMount _ = pure unit
 
-didUpdate :: Self -> Effect Unit
-didUpdate self = do
-  Console.log "did update payment select"
-  Console.log $ unsafeCoerce self.state
+didUpdate :: Self -> { prevProps :: Props, prevState :: State } -> Effect Unit
```
f-f: I guess this is an unrelated change that slipped in?

apps/mitt-konto/src/MittKonto/Main.purs
```diff
@@ -163,6 +164,12 @@ userView { user } = React.fragment
   , classy DOM.div "col col-12 md-col-6 lg-col-6" [ subscriptionsView ]
   ]
   where
+    subsSorted = foldl canceledLast [] user.subs
```
f-f: I wonder if we should start adding more docstrings, feels like some context would help coming back to this after some time

apps/prenumerera/src/Prenumerera/PaymentSelect.purs
vapaj: We should actually fix this in a separate PR.

packages/persona/src/Persona.purs
```diff
@@ -217,6 +224,13 @@ newtype SubscriptionState = SubscriptionState String
 derive instance genericSubscriptionState :: Generic SubscriptionState _
 instance readForeignSubscriptionState :: ReadForeign SubscriptionState where
   readImpl f = map SubscriptionState (readImpl f)
+instance eqSubscriptionState :: Eq SubscriptionState where
+  eq (SubscriptionState s1) (SubscriptionState s2) = s1 == s2
```
zudov: You should be able to derive it with:

```haskell
derive instance eqSubscriptionState :: Eq SubscriptionState
```

packages/persona/src/Persona.purs
```diff
@@ -217,6 +224,13 @@ newtype SubscriptionState = SubscriptionState String
...
+instance ordSubscriptionState :: Ord SubscriptionState where
+  compare s1 s2
+    | isSubscriptionStateCanceled s1 = GT
+    | isSubscriptionStateCanceled s2 = LT
+    | otherwise = EQ
```
zudov: This instance seems dodgy. It doesn't abides the laws either (see them [here](https://pursuit.purescript.org/packages/purescript-prelude/4.1.0/docs/Data.Ord#t:Ord)). If I understand correctly the intention was to make it so that cancelled subscription are always greater than others. I would have implemented it as:

```haskell
compare =
  comparing
    \s@(SubscriptionState st) ->
      if isSubscriptionStateCanceled s
      then Right st
      else Left st 
```

The idea here is that we defer all the real work to the existing `Ord (Either String String)` instance, so there can be no mistakes. We however implement the bit that we need by matching cancelled subscriptions and putting them as `Right` while everything else goes as `Left`, and [`Right` is always greater than `Left`](https://github.com/purescript/purescript-either/blob/v4.1.1/src/Data/Either.purs#L178)

packages/ksf-subscription/src/KSF/Subscription/Component.purs
```diff
@@ -56,8 +56,13 @@ render { props } =
...
     { product: props.subscription.package.name
     , status: translateStatus props.subscription.state
-    , nextBillingDate: trim $ fromMaybe "" $ formatDate =<< addOneDay props.subscription.dates.end
+    , nextBillingDate: nextBillingDate
     }
+  where
+    nextBillingDate
+      | Persona.isSubscriptionCanceled props.subscription = Nothing
+      | otherwise =
+          Just $ trim $ fromMaybe "" $ formatDate =<< addOneDay props.subscription.dates.end
```
zudov: Is there a need to do `fromMaybe ""` here? It could be just `map trim $ formatDate =<< addOneDay props.subscription.dates.end`.

apps/mitt-konto/src/MittKonto/Main.purs
```diff
@@ -181,9 +185,11 @@ userView { user } = React.fragment
       componentBlock "Mina prenumerationer:" $ subscriptions <> [ break, subscribeImage ]
       where
         subscriptions =
-          case user.subs of
+          case subsSorted of
             [] -> [ componentBlockContent noSubscriptionsText ]
-            subs -> map (componentBlockContent <<< subscriptionView) subs `snoc` cancelSubscription
+            subs -> do
+              let subscriptionViews = subscriptionView <$> subs
+              (subscriptionComponentBlockContent <$> zip subs subscriptionViews) `snoc` cancelSubscription
```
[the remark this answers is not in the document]
vapaj: Amazing point.

packages/ksf-subscription/src/KSF/Subscription/Component.purs
vapaj: Indeed

PR #52, opened 2019-01-02
Logout from Persona

vapaj:


f-f:
Note that we cannot merge this yet until we deploy the new version of Persona to production

PR #55, opened 2019-01-11
Expose the endpoint for gdpr updates in the login component

f-f:


vapaj:
Doesn't compile though

Issue #21, opened 2018-11-07
Add `registration` component


f-f:
Once we'll have the registration implemented in Persona, we should make a component for registration. Or maybe integrate the registration in the login component (since they always appear together anyways)

PR #58, opened 2019-01-24
Registration component

vapaj:
~Although it's still not quite finished, I guess it can be nonetheless reviewed.~
It's huge, sorry.

There is no js compatibility with this component atm.

Add #21


packages/ksf-login/src/KSF/Login/Component.purs
```diff
@@ -101,13 +103,16 @@ data Action =
   | FormPassword String
   | MergeCancel
   | SetMergeInfo (Maybe MergeInfo)
+  | SetViewStep View.LoginViewStep
```
vapaj: This is to set "steps" in the login/registration/user flow. I would have initially wanted to separate the account merging stuff from the login component and use it as a "step" as well.

packages/ksf-registration/src/KSF/Registration/Component.purs
```diff
@@ -0,0 +1,498 @@
+module KSF.Registration.Component where
```
vapaj: Note that I decided to only have a `Component` which has all the `View` stuff as well. I think the separation (at least in this case) makes things unnecessarily complex (most of this code is view related anyway).

packages/ksf-registration/src/KSF/Registration/Component.purs
```diff
@@ -0,0 +1,500 @@
...
+  UpdateInput Country newValue ->
+    Update self.state { country = newValue }
+  UpdateInput Phone newValue ->
+    Update self.state { phone = newValue }
+  UpdateInput EmailAddress newValue ->
+    Update self.state { emailAddress = newValue }
+  UpdateInput Password newValue ->
+    Update self.state { password = newValue }
+  UpdateInput ConfirmPassword newValue ->
+    Update self.state { confirmPassword = newValue }
```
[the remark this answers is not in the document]
vapaj: `FirstName -> { firstName = newValue }` this is not allowed syntax, however

```haskell
UpdateInput attr newValue -> Update do
  case attr of 
    FirstName -> self.state { firstName = newValue }
    LastName  -> self.state { lastName = newValue }
...
```

would still be nicer.

PR #63, opened 2019-03-20
Elections

joakimekroos:
Set up project


zudov:
I tried to build it and it turned out that the top-level `.gitignore` excludes `apps/*/index.js`, so I've unexcluded it. Works for me now :D

PR #67, opened 2019-03-24
Load candidates table data from the backend API

zudov:
Currently just hardcodes fetching of area `001` as we don't have a selector.

### DIFF
--- a/apps/elections/components/Table.js
+++ b/apps/elections/components/Table.js
@@ -60,10 +60,10 @@ export default class Table extends React.Component {
                 Header: 'Status',
                 id: 'status',
                 accessor: ({status}) =>
-                    status === Status.ELECTED     ? "Elected"
-                  : status === Status.SUBSTITUTE  ? "Substitute"
-                  : status === Status.NOT_ELECTED ? "Not Elected"
-                  : status === Status.INELIGIBLE  ? "Ineligible"
+                    status === Status.ELECTED     ? "Invald"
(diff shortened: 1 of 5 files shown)


apps/elections/components/Table.js
```diff
@@ -34,23 +42,29 @@ export default class Table extends React.Component{
...
               },
               {
                 Header: 'Status',
-                accessor: 'status'
+                id: 'status',
+                accessor: ({status}) =>
+                    status === Status.ELECTED     ? "Elected"
+                  : status === Status.SUBSTITUTE  ? "Substitute"
+                  : status === Status.NOT_ELECTED ? "Not Elected"
+                  : status === Status.INELIGIBLE  ? "Ineligible"
```
zudov: Please translate these to Swedish before merging.

apps/elections/components/Table.js
```diff
@@ -3,26 +3,34 @@ import { render } from 'react-dom';
...
+                accessor: 'candidateNumber'
               },
               {
                 Header: 'Parti',
-                accessor: 'party'
+                // TODO: when hovering over that cell, we could show a box with extended party info
+                id: 'nominator.abbreviation',
+                accessor: ({ nominator }) =>
+                  // ???: Self-nominated candidates have 'E<candidateNumber>' abbreviation
+                  //      We could render some text instead of that.
```
zudov: What do you think about this?

PR #79, opened 2019-04-10
Hide some cols on mobile.

joakimekroos:
Added react-device-detect
Hidden some cols for mobiles
Changed order of first name and last name


apps/elections/components/Table.js
```diff
@@ -12,6 +13,7 @@ export default class Table extends React.Component {
   render() {
     const { areaResponse } = this.props;
     const data = areaResponse === null ? null : areaResponseData(areaResponse);
+    // console.log("isMobile: ", isMobile);
```
f-f:
```suggestion
```

PR #97, opened 2019-04-15
Pause subscription

vapaj:
Adds pausing of subscriptions.

TODO:

- [x] send calls to Persona
- [x] handle success/error


packages/ksf-pause-subscription/src/KSF/PauseSubscription/Component.purs
```diff
@@ -0,0 +1,233 @@
...
+              , maxDate: toNullable $ fromDateTime <$> maxDate
+              , disabled
+              , locale: "sv-SV"
+              } ]
+    ]
+    $ Just { extraClasses: [ "mt2" ] }
+
+submitForm :: State -> Props -> Effect Unit
+submitForm { startDate: Just start, endDate: Just end } props@{ userUuid, subsno } = do
+  loginResponse <- Login.loadToken
```
f-f: 🤔 I wonder if we should avoid leaking these details into other components, and have the `Login` component should handle authenticated Persona calls (and at that point we should rename it to `User` component to make it clear).

Another option is to handle all the login-related stuff into a the `Persona` component, but that might be harder

vapaj: Good point, a `User` component makes a lot of sense.

less/ksf-grid.less
```diff
@@ -0,0 +1,9 @@
+@import (reference) '../less/ksf-variables';
+
+.grid--row {
```
f-f: Empty rule?

packages/persona/src/Persona.js
```diff
@@ -5,7 +5,7 @@ Persona.ApiClient.instance.basePath = process.env.PERSONA_URL;
 
 // https://visionmedia.github.io/superagent/#timeouts
 Persona.ApiClient.instance.timeout = {
-  response: 5000, // the server only has 5 seconds to respond
+  response: 50000, // the server only has 5 seconds to respond
```
f-f: I think this wasn't supposed to be committed 😄

packages/ksf-pause-subscription/src/KSF/PauseSubscription/Component.purs
```diff
@@ -0,0 +1,233 @@
...
+  SetMinStartDate newMinStartDate -> self.state { minStartDate = newMinStartDate }
+
+-- | Minimum pause period is one week
+calcMinEndDate :: Maybe DateTime -> Maybe DateTime
+calcMinEndDate Nothing = Nothing
+calcMinEndDate (Just startDate) = do
+  let week = Time.Duration.Days 7.0
+  adjust week startDate
+
+-- | Maximum pause period is three months
```
f-f: Do we have business rules for how much time there has to be between two pauses?

packages/ksf-pause-subscription/src/KSF/PauseSubscription/Component.purs
```diff
@@ -0,0 +1,233 @@
...
+  SetStartDate newStartDate ->
+    self.state
+      { startDate = newStartDate
+      , minEndDate = calcMinEndDate newStartDate
+      , maxEndDate = calcMaxEndDate newStartDate
+      }
+  SetEndDate newEndDate -> self.state { endDate = newEndDate }
+  SetMinStartDate newMinStartDate -> self.state { minStartDate = newMinStartDate }
+
+-- | Minimum pause period is one week
```
f-f: AFAIK we don't have these checks on the backend right? If no, we should ship them before this gets merged

less/ksf-grid.less
vapaj: There will probably be some rules at some point, although for now it can be removed

packages/persona/src/Persona.js
vapaj: Nice catch :D Indeed it wasn't

packages/ksf-pause-subscription/src/KSF/PauseSubscription/Component.purs
vapaj: Good question. There should be at least one week in between pauses (cc @kirstipa). Although this is not easy to show for the user, as the _amazing_ date picker does not support disabling specific dates. A proper error message goes a long way though.

packages/ksf-pause-subscription/src/KSF/PauseSubscription/Component.purs
vapaj: The Persona errors are now handled nicely 🙌

PR #135, opened 2019-07-02
Refactor resources

vapaj:


.gitignore
```diff
@@ -9,8 +9,6 @@ package-lock.json
 .DS_Store
 
 packages/*/index.js
-apps/*/index.js
-!apps/elections/index.js
```
f-f: Let's explicitly enumerate the apps that should have their index ignored, instead of special casing the JS ones

apps/mitt-konto/package.json
```diff
@@ -13,7 +13,6 @@
   },
   "scripts": {
     "start": "parcel index.html -p 8000",
-    "build": "yarn build-purs && parcel build index.html --no-cache && cp xd_receiver.html ./dist",
-    "build-purs": "cd ../.. && spago bundle-module -m MittKonto.Main -t apps/mitt-konto/index.js --path \"packages/*/src/**/*.purs\" --path \"apps/*/src/**/*.purs\" "
+    "build": "parcel build index.html --experimental-scope-hoisting --no-cache && cp xd_receiver.html ./dist"
```
f-f: We'd still need a purs-building step here right? Adding a `spago build` should be a no-op if you already built, and will update the generated js if you didn't

deploy.rb
```diff
@@ -69,6 +69,7 @@ def run_command(command)
 build_commands = [
   "yarn run clean",
   "yarn install --pure-lockfile --cache-folder=.yarn-cache",
+  "spago build",
```
f-f: And if we include the `spago build` in the apps themselves this should not be needed

apps/mitt-konto/package.json
vapaj: Yeah, maybe better have it here

deploy.rb
vapaj: Rather than here

PR #136, opened 2019-07-03
Refactor Spago projects

f-f:
This PR should complete the build-system refactoring started in #135 

Most relevant changes:
- upgrade to PureScript 0.13
- most of the components now have been merged in a single `components` library (exception for now is the `login` that needs to be a standalone package, and soon things like `navbar`, etc)
- each project that is a PureScript library or is somehow a PureScript build now has a `spago.dhall` configuration (both `apps` and `packages`, for an example [see this one](https://github.com/KSF-Media/affresco/pull/136/files#diff-e899317ff730e7619c43631b3105d2a5))
  Note how in the linked example we depend on things like `affresco-login`
- We can do this because there is only one shared `packages.dhall` in the top-level, where we specify libraries that are local to the repo, by just importing their `spago.dhall as Location` (this is to avoid circular imports in Dhall), e.g. see [here](https://github.com/KSF-Media/affresco/pull/136/files#diff-95c189e71349444d37d671bcc7a0de60R133)
- another cool thing here is that we can have granular dependencies and have the libraries nicely depend on each other, e.g. see how the login library depends on the `components` [here](https://github.com/KSF-Media/affresco/pull/136/files#diff-f37b5e4a2d6dc00d0856c70f0db9e88cR20)


f-f:
Note: CI is not passing because we need https://github.com/spacchetti/spago/pull/301 to be merged and a new release of Spago to be released first

PR #144, opened 2019-07-16
remove share and sound icons

lulzimfazlija:


apps/app-article/src/components/article-additional.js
```diff
@@ -23,23 +20,24 @@ class Additional extends Component {
...
             <div className={"row"}>
                 <div className="col-8">
-                    {
-                        <a href="javascript:void(0)" onClick={() => {
-                            this.onShare()
-                        }} id={"shareLink"}>
-                            <img style={{marginTop: '4px'}} className={"shareIcon"} src={shareIcon}
-                                 alt="Share"/></a>
-                    }
+                    {/*{*/}
```
f-f: Could we remove this code instead of commenting it? Once we want to add it back we can always come back to this Pull Request

lulzimfazlija: sure yeah :)

PR #147, opened 2019-07-29
Require JanrainSSO.Config in all the functions that call to SSO

zudov:
This way it's impossible to call those functions without having a
config, and the only way to get a config is by calling a function
`loadConfig :: Effect (Maybe Config)` which will ensure that the
library is loaded.


f-f:
There seems to be something weird with the CI, the PureScript build is not supposed to run on JS-only projects 🤔

f-f:
I'll try to merge #136 (which should fix all CI errors) today so we can rebase this on that one

zudov:
OK. Full disclosure: I haven't try to build it locally yet, so it might as well fail to compile.

f-f:
@zudov no worries 🙂 #136 needs to be merged anyways, but you might want to try to build locally in the meanwhile 😄

PR #154, opened 2019-08-19
Refactor navbar

vapaj:
- simplify layout
- fix layout bugs


packages/components/src/Navbar/Component.purs
```diff
@@ -66,37 +70,95 @@ navbar = make component
...
+-- | Items of hamburger nav
+collapsedNav :: Self -> JSX
+collapsedNav self =
+  Collapsed.collapsed
+    { visibility: self.state.collapsedNavVisibility
+    , navItems: [ logoutButton self, needHelp self.props.paper ]
+    }
+
+onLogout :: Self -> Persona.User -> Maybe (Effect Unit)
+onLogout self _user = Just do
```
zudov: Let's move the `Maybe` out of here and have the `Just onLogout` at the call sites, if it's actually needed.

packages/components/src/Navbar/Component.purs
```diff
@@ -66,37 +70,95 @@ navbar = make component
...
-send = runUpdate update
+hamburgerButton :: Self -> JSX
+hamburgerButton self =
+  DOM.div
+    { className: "nav--hamburger-button"
+    , onClick: Event.handler_ $ self.setState _ { collapsedNavVisibility = negateVisibility self.state.collapsedNavVisibility }
+    , children:
+        [ hamburgerStripe
+        , hamburgerStripe
+        , hamburgerStripe
```
zudov: Possible, but not necessarily better to do `replicate 3 hamburgerStripe`

PR #159, opened 2019-08-30
removed ls from everywhere

wiltso:
Think i have removed ls from everywhere


apps/mitt-konto/package.json
```diff
@@ -7,6 +7,7 @@
     "basscss": "^8.0.4",
     "create-react-class": "^15.6.3",
     "less": "^3.8.1",
+    "node-gyp": "^5.0.3",
```
vapaj: This was probably added while attempting to get node+purescript+parcel up and running 😄
It's not needed, so go ahead and remove this.

Note: this is kind of a two part thing, because when doing `yarn install` with new packages, the `yarn.lock` file is changed as well. So that file should not be modified either.

Removing this line and then running `yarn install` should restore the `yarn.lock` to its previous state.

packages/components/src/Icons.js
```diff
@@ -2,7 +2,6 @@
 exports.papers = {
   hbl: require('../../../../images/logo-hbl.svg'),
   hthl: require('../../../../images/logo-hthl.svg'),
-  ls: require('../../../../images/logo-ls.svg'),
```
vapaj: Let's also remove the actual `logo-ls.svg` file found in `images/`.

PR #161, opened 2019-09-02
User component

vapaj:


packages/user/src/User/User.purs
```diff
@@ -0,0 +1,361 @@
+module KSF.User
+  ( UserError (..)
+  , MergeInfo
+  , ValidationServerError
+  , module PersonaReExport
```
f-f: Does this reexport only the things that we imported in scope or the whole module? If the latter, why reexport it? (since the point of the `User` component is to shield from implementation details of the login)

vapaj: It exports only the things we import:
```
import Persona (User, MergeToken, Provider(..), UUID, Email(..), Token(..), InvalidPauseDateError(..), InvalidDateInput(..), UserUpdate(..)) as PersonaReExport
```

f-f: Awesome, thanks 🙂

PR #166, opened 2019-09-05
text changes, layout improve

lulzimfazlija:


apps/app-article/src/article-service.js
```diff
@@ -11,7 +11,7 @@ const articleApi = {
 
     },
     getLatestArticles() {
-        return fetch(config.apiUrl + "latest?start=0&limit=5", {
+        return fetch(config.apiUrl + "mostread?start=0&limit=5", {
```
f-f: Is this right? I mean, the function is called `getLatestArticles`, but we fetch the `mostread` endpoint

apps/app-article/src/App.js
```diff
@@ -463,8 +469,7 @@ class App extends Component {
                                 {
                                     this.state.appearLogin ?
                                         <Login onRegister={() => this.onRegisterOpen()}
-                                               onUserFetchSuccess={(user) => this.onUserFetchSuccess(user)}
-                                               disableSocialLogins={[]}/>
```
f-f: Let's leave this `disableSocialLogins` in, as a reminder for the future (we did disable some things in the last days)

apps/app-article/src/App.js
```diff
@@ -389,6 +389,12 @@ class App extends Component {
             articleTitle.style.lineHeight = "100%";
         }
 
+        if (document.getElementsByClassName('preamble').length > 0) {
```
vapaj: Just a note: When using React, it's not advisable to directly manipulate the DOM (there is a possibility when `setState` is called, changes made like this are overridden). It would be better to, somehow, resize the fonts by making a `state` change (in combination of using css).

lulzimfazlija: You are right @vapaj , I will consider rewriting the increaseFontSize functionality next release! I guess it is going to be changed anyhow and some of its funcionality be managed by the native side!

apps/app-article/src/article-service.js
```diff
@@ -3,15 +3,21 @@ import {isUserLoggedIn} from "./helper";
...
+        return fetch(config.production.apiUrl + "article/" + uuid, {
             method: 'GET',
             headers: attachHeaders()
         })
             .then(response => response.json())
 
     },
     getLatestArticles() {
-        return fetch(config.apiUrl + "latest?start=0&limit=5", {
+        return fetch(config.production.apiUrl + "latest?start=0&limit=5", {
```
vapaj: This should actually be `config.apiUrl`. If you look at https://github.com/KSF-Media/affresco/blob/master/apps/app-article/src/config.js, I've removed the `production` and `dev` keys there, and use `.env` to figure out which url to use (this way, it's in sync with our other deploys and workflow).

Note: the default dev `LETTERA_URL` is added here https://github.com/KSF-Media/affresco/blob/master/apps/app-article/.env

PR #169, opened 2019-09-16
application banner js code added

jockeF:


scripts/app/gamAppAdsMin.js
```diff
@@ -0,0 +1 @@
...="#" title="stäng annons" onclick="return ksfDfp.closeInterstitial(\''+contentUnitDiv+"');\">x</a></p></header>")}}})});ksfDfp.closeInterstitial=function(interstitialDivName){var divToClose=document.getElementById(interstitialDivName).parentNode;document.body.className=document.body.className.replace("noScroll","");divToClose.style.display="none";return false};ksfDfp.closeSlotAfterCallback=function(slot){var divToClose=document.getElementById(slot);divToClose.style.display="none";var headerToClose=document.getElementsByClassName("ksfDFPadHeader "+slot);if(headerToClose[0]!=="undefined"){headerToClose[0].style.display="none !important"}};ksfDfp.allowedPostMessages=["closeSlot"];ksfDfp.createBannerOnMessage=function(e){if(!(e.data&&ksfDfp.allowedPostMessages.indexOf(e.data.cmd)!==-1)){return}var message=e.data;var bannerIdString=message.id;switch(message.cmd){case"closeSlot":ksfDfp.closeSlotAfterCallback(bannerIdString);break;default:return}};if(window.addEventListener){window.addEventListener("message",ksfDfp.createBannerOnMessage,false)}else{if(window.attachEvent){window.attachEvent("onmessage",ksfDfp.createBannerOnMessage)}else{window.onmessage=ksfDfp.createBannerOnMessage}}}})();
```
vapaj: I don't think we need the minified version in the repo.

jockeF: Agree. Will remove it.

jockeF: **Some brief documentation for the script**
Requirements:
The gpt.js from Google
<script async='async' src='https://www.googletagservices.com/tag/js/gpt.js'></script>

The script itself from where ever it is hosted
<script async='async' src='gamAppAdsMin.js'></script>

DIV tags in the HTML that correspond to the once known by the script:
The script is currently looking for div:s in the HTML, with an id among these:
MOBPARAD
MOBBOX1
MOBBOX2
MOBBOX5
These names are not final. The expected HTML is this:
<div id = “string”></div>

/Requirements

GPDR consent is hard-coded as using the app will require consent.

The code listens for ads marked 
newspaper > app
and there is an order set up in Google Ad Manager to populate the app during testing.

In production there will presumably be three values for newspaper, one for each paper.

The div tags above all currently support these sizes: 
300*100, 300*250, 300*300, 300*600
The sizes are also not final.

OnSwitch is hard-coded as true at the moment. Should probably be set according to something it can listen for. There is a requirement for this functionality but it cannot be provided by Google Ad Manager. So it has to be set here.

f-f: @jockeF thanks for the docs! Could we move them to a markdown file near the script so that they stay versioned with the file?

jockeF: @f-f Yes, that makes sense. This description though is very much a temp file :-). But sure, will add it and update as it matures.

vapaj:
Wonderful! 👏

PR #179, opened 2019-10-07
store user token into coockies and recieve from coockies when making …

lulzimfazlija:
…request to the backend, this change is done due the native android not able to read localstorage


apps/app-article/src/helper.js
```diff
@@ -1,11 +1,16 @@
...
-    return !(localStorage.getItem("uuid") === null && localStorage.getItem("token") === null);
+    return !(getCookie('token') === null);
 };
 
 export const shareArticle = (title, url, description) => {
     return AndroidNativeShare(title, url, description);
 };
 
+export const getCookie = (name) => {
+    let v = document.cookie.match('(^|;) ?' + name + '=([^;]*)(;|$)');
```
vapaj: This is a bit hard to follow, what's going on here? 🙂 Why do we need three capture groups?

vapaj:
As the `User` component will automatically clear the local storage if the `token` is expired, we probably should manually remove the cookie when this happens. Doing this in the `onUserFetchFail` callback should do the trick.

You can see all the available callbacks here https://www.npmjs.com/package/@ksf-media/user

apps/app-article/src/App.js
```diff
@@ -68,10 +68,23 @@ class App extends Component {
             this.setState({fontSize: parseFloat(localStorage.getItem("fontSize"))});
         }
 
+        if(getCookie('LoggedOutFlag')){
+            let d = new Date();
+            d.setTime(d.getTime() - (1000*60*60*24)); //Set the time to the past. 1000 milliseonds = 1 second
+            let expires = "expires=" + d.toUTCString();
+            window.document.cookie = "LoggedOutFlag=" + "; "+expires;
+            logout(this.onLogout);
+            localStorage.removeItem('token');
```
vapaj: Calling `User.logout` will clean up the local storage from `uuid` and `token`, so you don't need to do this.

apps/app-article/src/App.js
```diff
@@ -375,6 +391,10 @@ class App extends Component {
         this.fetchArticleFromApi(urlParams.get('uuid'));
     }
 
+    onUserFetchFail(error){
```
vapaj: What will happen here when you don't set/remove any cookies? `onUserFetchFail` is most likely invoked because the `token` is expired (this means the user is in fact logged out). How will this effect the native part?

apps/app-article/src/App.js
```diff
@@ -113,14 +126,17 @@ class App extends Component {
...
-            const articleFound = JSON.parse(localStorage.getItem('cachedArticles')).find(article => {
-                if (article.uuid === uuid) {
-                    return true;
-                }
-            });
-            return articleFound;
+        if(!isUserLoggedIn()){
+            if (JSON.parse(localStorage.getItem('cachedArticles')) != null) {
+                const articleFound = JSON.parse(localStorage.getItem('cachedArticles')).find(article => {
+                    if (article.uuid === uuid) {
```
vapaj: This could be simply `return article.uuid === uuid`.

apps/app-article/src/App.js
```diff
@@ -113,14 +125,15 @@ class App extends Component {
...
     checkCache(uuid) {
-        if (JSON.parse(localStorage.getItem('cachedArticles')) != null) {
-            const articleFound = JSON.parse(localStorage.getItem('cachedArticles')).find(article => {
-                if (article.uuid === uuid) {
-                    return true;
-                }
-            });
-            return articleFound;
+        if(!isUserLoggedIn()){
+            if (JSON.parse(localStorage.getItem('cachedArticles')) != null) {
```
vapaj: We should first check if we find the article in the local storage, and then `JSON.parse` only one time.

apps/app-article/src/App.js
```diff
@@ -68,10 +68,22 @@ class App extends Component {
             this.setState({fontSize: parseFloat(localStorage.getItem("fontSize"))});
         }
 
+        if(getCookie('LoggedOutFlag')){
+            let d = new Date();
+            d.setTime(d.getTime() - (1000*60*60*24)); //Set the time to the past. 1000 milliseonds = 1 second
+            let expires = "expires=" + d.toUTCString();
```
vapaj: Also, should we move the cookie expiration inside the `onLogout` callback?

vapaj: Because it would be correct to do it after the logout occurs, not actually before.

apps/app-article/src/helper.js
```diff
@@ -6,6 +6,13 @@ export const shareArticle = (title, url, description) => {
     return AndroidNativeShare(title, url, description);
 };
 
+export const getCookie = (name) => {
```
vapaj: Let's remove this as it seems it's not used anywhere

apps/app-article/src/App.js
```diff
@@ -369,10 +394,21 @@ class App extends Component {
     }
 
     onUserFetchSuccess(user) {
+        //Cookie will expire after 7 days 
+        Cookies.set('LoginStatus', true, { expires: 7 });
+        //To get User data from Android side 
+        Cookies.set('currentUser', JSON.stringify(user), { expires: 7 });
```
vapaj: The `user` might be quite large. Do we need everything from there?

apps/app-article/src/App.js
```diff
@@ -68,23 +72,38 @@ class App extends Component {
...
             this.setState({cachedArticles: JSON.parse(localStorage.getItem('cachedArticles'))});
         }
-
-        let urlParams = new URLSearchParams(window.location.search);
-        if (urlParams.has('uuid')) {
+        if (getUrlParam().has('uuid')) {
             if (this.checkCache(urlParams.get('uuid'))) {
                 this.fetchArticleFromCache(urlParams.get('uuid'));
             } else {
                 this.fetchArticleFromApi(urlParams.get('uuid'));
```
mrskug: where do we get urlParams from as the declaration is removed?

baderkhane: @mrskug fixed now, thanks for remark :)

PR #185, opened 2019-10-23
Fix/dullen

wiltso:


apps/duellen/app/components/Intro.jsx
```diff
@@ -66,14 +66,46 @@ export default class Intro extends React.Component{
...
+              weeknum = Math.floor((daynum+day-1)/7);
+          }
+          return weeknum;
+      };
+      var date = this.state.quizData.publication_date
+      var mydate = new Date(Number(date.slice(0,4)), Number(date.slice(5,7))-1,Number(date.slice(8,10)));
+      this.setState({
+        week: mydate.getWeek(),
+      });     
+  }
```
zudov: The same `getWeek` appears both in `Start.jsx` and `Intro.jsx`. I'd recommend to place it in a separate file and importing it in both places. Or maybe there is an npm package for this.

PR #192, opened 2019-10-29
Added bootstrap to the hole front end to make it more responsive on m…

wiltso:
…obile.


apps/duellen/app/components/Question.jsx
```diff
@@ -106,11 +110,20 @@ export default class Question extends React.Component {
...
     }else{
-      this.setState({check: 'Fel!', opacity: 1, color: red,}, () => setTimeout(() => this.setState({check: '', opacity:0}),3000));  
+      console.log(hintPoint)
+      if(hintPoint === 1){
+      this.setState({
+        check: 'Nu var det dags för nästa fråga, skriv in svaret här',
+        color: "white",
+      });
+      }else{
+        console.log("UIYIKFYLGLIUOI")
```
f-f: I guess this was not supposed to get in 😄

apps/duellen/index.html
```diff
@@ -3,98 +3,21 @@
...
-    </div>
+    <main class="container mt-5">
+      <div id="root">
+    
+      </div>
+    </main>
     <script type="text/javascript" src="index.js"></script>
+    <!-- Optional JavaScript -->
+    <!-- jQuery first, then Popper.js, then Bootstrap JS -->
+    <script src="https://code.jquery.com/jquery-3.2.1.slim.min.js" integrity="sha384-KJ3o2DKtIkvYIK3UENzmM7KCkRr/rE9/Qpg6aAZGJwFDMVNA/GpGFF93hXpG5KkN" crossorigin="anonymous"></script>
```
f-f: We should add these scripts via `npm` instead of importing them like this, otherwise offline development wouldn't work

PR #193, opened 2019-10-30
Wikilinks

wiltso:


apps/duellen/app/components/Question.jsx
```diff
@@ -1,6 +1,7 @@
 import React from 'react';
 import AutoComplete from 'material-ui/AutoComplete';
 import MuiThemeProvider from 'material-ui/styles/MuiThemeProvider';
```
f-f: We should probably remove this is we're not using it anymore

apps/duellen/app/components/WikiLink.jsx
```diff
@@ -0,0 +1,64 @@
+import React from 'react';
+import $ from 'jquery';
+import {withRouter} from 'react-router';
+import MuiThemeProvider from 'material-ui/styles/MuiThemeProvider';
```
f-f: Same here

PR #197, opened 2019-11-01
Made a new funkction so if you skip a question you get a message that…

wiltso:
… says you skipt this question and added so the messages won't stack on each other

### DIFF
--- a/apps/duellen/app/components/Question.jsx
+++ b/apps/duellen/app/components/Question.jsx
@@ -106,18 +106,18 @@ export default class Question extends React.Component {
     const {tally, hintPoint} = this.state;
     if(this.state.userInput === this.checkIfCorrect()){
       this.hideMessage()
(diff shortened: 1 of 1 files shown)


apps/duellen/app/components/Question.jsx
```diff
@@ -91,23 +91,43 @@ export default class Question extends React.Component {
...
+  };
 
   handleWrongRight(e){
     e.preventDefault();
     const {tally, hintPoint} = this.state;
     if(this.state.userInput === this.checkIfCorrect()){
-      cogoToast.success('Du fick den rätt');
-      this.setState({tally: tally + hintPoint});
+      this.hideMessage()
+      var test = cogoToast.success('Du fick den rätt', {toastContainerID: '1'});
```
vapaj: Let's rename the `test` variable to something more meaningful (unless it really is a test of some sort :D)

PR #229, opened 2019-11-20
View past hints

wiltso:


apps/duellen/app/components/Question.jsx
```diff
@@ -136,9 +141,16 @@ export default class Question extends React.Component {
     // Sets the user input to nothing so the input field is empty
     this.setState({
       userInput: '',
+      pastQuestionHints: [],
     })
   };
 
+  showHints(){
+    console.log(!this.state.showPastHints)
```
vapaj: Let's remove debug logging here

PR #233, opened 2019-11-25
Pagination

wiltso:


apps/duellen/app/components/Start.jsx
```diff
@@ -29,21 +36,27 @@ export default class Start extends React.Component {
      try {
        // The backendUrl needs to be the same as from where you want the quizzes to be fetched
        // 'get/all/quizzes/as/json' can't have a / in the end because the backend address dosen't have it
-       const res = await fetch(backendURL + 'get/all/quizzes/as/json');
+       const res = await fetch(backendURL + 'get/all/quizzes/as/json/?lowerlim=' + this.state.lowerLim.toString() + '&upperlim='+ this.state.gap.toString());
        const quizData = await res.json();
+       console.log(quizData)
```
vapaj: Let's remove logging here

apps/duellen/app/components/Start.jsx
```diff
@@ -58,32 +71,60 @@ export default class Start extends React.Component {
     this.props.history.push('/Intro');
   }
 
+  // Loads more quizzes
+  loadMore(e){
+    e.preventDefault();
+    try {
+      // Gets the more quizzes from the backend        
+      fetch(backendURL + 'get/all/quizzes/as/json/?lowerlim=' + this.state.lowerLim.toString() + '&upperlim='+ this.state.upperLim.toString()
+          ).then(
```
vapaj: Ah, sorry for missing this when you asked me :D But I guess usually the indentation/new lines are done like this
```
fetch(...)
.then(...)
.then(...)
```
But I'm fine with this.

vapaj: We should really use a linter...

apps/duellen/app/components/Start.jsx
```diff
@@ -58,32 +70,58 @@ export default class Start extends React.Component {
...
                 {this.state.quizData.map(item => (
-                  <div key={item.id} class="btnOrange">
+                  <div key={item.id} className="btnOrange">
                     <a href={'/Intro/' + item.id}>
-                      <Button variant="contained" class="start" color="primary" fullWidth={true}>Vecka {this.getweek(item.publication_date)}</Button>
+                      <Button variant="contained" className="start" color="primary" fullWidth={true}>Vecka {this.getweek(item.publication_date)}</Button>
                     </a>
                   </div>
                 ))}
+                {this.state.showingAll == false &&
```
vapaj: It would be better to have this as `this.state.showingAll === false && ...`.
Look https://codeburst.io/javascript-double-equals-vs-triple-equals-61d4ce5a121a for context

PR #241, opened 2019-12-19
Meeting fixes

wiltso:
Change the logo so it's the right year on the logo
Shows the amount of points you got from the question on the middle page
Shows the correct text of how you did for each question at the end
Comented out the description from the wiki links so it's easy to put back if needed
And some small general styling changes


apps/duellen/app/components/Intro.jsx
```diff
@@ -27,6 +27,7 @@ export default class Intro extends React.Component{
     try {
       const res = await fetch(backendURL + 'get/all/quizzes/as/json/' + this.props.match.params.id);
       const quizData = await res.json();
+      console.log(quizData)
```
vapaj: Let's remove logging here

apps/duellen/app/components/WikiLink.jsx
```diff
@@ -47,19 +47,18 @@ export default class WikiLink extends React.Component{
           <div className="w-100">
             {this.state.wikiresponse.map(({title, description}) => (
               <button key={title} onClick={() => this.props.onClick(title)} type="button" className="w-100 p-3 mb-2 bg-light text-black border">
-                  <p>
                     <b>{title}</b>
+                    {/* Test if it should only show the title
```
vapaj: What's this test? 😀

wiltso: Steffen wanted to test with some users if it's better with just the title

vapaj: Should we remove the commented stuff?

PR #252, opened 2020-01-20
Add hashes of purescript dependencies

vapaj:
Add build hashes of purescript projects, so that we are able to successfully setup the [Netlify monorepo](https://www.netlify.com/blog/2019/10/09/launching-monorepo-support-for-netlify-sites/) deploys. The hashes are checked on `pre-commit` hook.


hooks/pre-commit
```diff
@@ -0,0 +1,75 @@
...
+    foreach my $dir (@APP_DIRECTORIES) {
+	my $full_app_dir = "$CWD/$dir";
+	opendir(DH, $full_app_dir)
+	    or die "Could not open directory: '$full_app_dir'";
+
+	# Is "output" directory found?
+	if (grep(/^output$/, readdir DH)) {
+	    chdir $full_app_dir;
+	    # Calculate hash of build directory and write to file if changed
+	    my $build_hash_output = `find . -type f | xargs echo | sha256sum`;
```
zudov: `xargs echo` :cat:

vapaj: also `find ./output` 🤠

PR #256, opened 2020-02-03
check for null, show default img instead

lulzimfazlija:


apps/app-article/src/components/card.js
```diff
@@ -31,7 +31,13 @@ const Card = (props) => {
                         <div className={"article-premium-badge"}><PremiumBadge/></div>
                         : ''
                 }
-                <img className={"card-img-top"} src={props.article.listImage.url + "&function=hardcrop&width=798&height=649&q=95"}/>
+                {
+                    props.article.listImage.url === null 
+                        ? <img className={"card-img-top"} src={props.article.listImage.url + "&function=hardcrop&width=798&height=649&q=95"}/>
```
vapaj: This should probably be the `hblDefaultImage`, because `props.article.listImage.url === null` 🙂

lulzimfazlija: true, i realized it after I comited :)

apps/app-article/src/components/card.js
```diff
@@ -31,7 +32,13 @@ const Card = (props) => {
                         <div className={"article-premium-badge"}><PremiumBadge/></div>
                         : ''
                 }
-                <img className={"card-img-top"} src={props.article.listImage.url + "&function=hardcrop&width=798&height=649&q=95"}/>
+                {
+                    props.article.listImage.url != null
```
vapaj: Let's use `!!props.article.listImage.url ? ... : ...` here, as this will prevent `null`, `undefined` and empty strings to be evaluated true.

lulzimfazlija: thank you, this is way better!

Issue #242, opened 2020-01-03
Delivery reclamation


ae-mo:
We should expose an interface that lets the user create delivery reclamations through mitt konto.

ae-mo:
Not only that, we should create an app for the customer support as well.

PR #257, opened 2020-02-05
Feature/delivery reclamation

ae-mo:
First version of #242. 

Needs proper layout and styles at the very least, also Swedish texts.

### DIFF
--- a/less/DeliveryReclamation.less
+++ b/less/DeliveryReclamation.less
@@ -2,22 +2,8 @@
 
 .delivery-reclamation--container {
     h3 { margin-top: 0 };
-    input[type=text] {
-	box-sizing: border-box;
-	width: 100%;
-    }
     justify-content: flex-end;
 
-    input[type=text] {
-	width: 100%;
-	font-size: 1em;
-	border-style: solid;
-	border-width: 1px;
-	border-color: @color-input-border;
-	padding: 8px 10px 8px;
-	background: @white;
-	box-sizing: border-box;
-    }
 }
 
 .delivery-reclamation--close-icon {
--- a/less/DeliveryReclamation.less
+++ b/less/DeliveryReclamation.less
@@ -1,0 +1,40 @@
+@import (reference) 'ksf-colors';
+
+.delivery-reclamation--container {
+    h3 { margin-top: 0 };
+    input[type=text] {
+	box-sizing: border-box;
+	width: 100%;
+    }
+    justify-content: flex-end;
+
+    input[type=text] {
+	width: 100%;
+	font-size: 1em;
+	border-style: solid;
+	border-width: 1px;
+	border-color: @color-input-border;
+	padding: 8px 10px 8px;
+	background: @white;
+	box-sizing: border-box;
+    }
+}
+
+.delivery-reclamation--close-icon {
+    justify-content: right;
+    align-items: center;
+    height: 1.5em;
+    .close-icon {
+	cursor: pointer;
+    }
+}
+
+.delivery-reclamation--date-picker {
+    width: 100%;
+    .react-date-picker__wrapper {
+	width: 100%;
+	border-color: @grey-mediumlight;
+	border-radius: 3px;
+	padding: 3px;
+    }
+}
--- a/less/InputField.less
+++ b/less/InputField.less
@@ -39,6 +39,24 @@
     }
 }
 
+.input-field--radio-container {
+  display: grid;
+  grid-template-columns:
+    [radio-button] 2em
+    [label] auto;
+  input[type=radio] {
+    grid-column: radio-button;
+    grid-row: 1; /* NEW */
+  }
+  label {
+    grid-column: label;
+    grid-row: 1; /* NEW */
+  }
+  grid-column-gap: 20px;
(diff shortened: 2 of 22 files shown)


less/DeliveryReclamation.less
```diff
@@ -0,0 +1,40 @@
+@import (reference) 'ksf-colors';
+
+.delivery-reclamation--container {
+    h3 { margin-top: 0 };
+    input[type=text] {
```
vapaj: I don't think we have any text inputs here, so this can be removed.

less/DeliveryReclamation.less
```diff
@@ -0,0 +1,40 @@
...
+	box-sizing: border-box;
+	width: 100%;
+    }
+    justify-content: flex-end;
+
+    input[type=text] {
```
vapaj: Also here

less/InputField.less
```diff
@@ -39,6 +39,24 @@
...
 }
 
+.input-field--radio-container {
+  display: grid;
+  grid-template-columns:
+    [radio-button] 2em
+    [label] auto;
+  input[type=radio] {
+    grid-column: radio-button;
+    grid-row: 1; /* NEW */
```
vapaj: Let's remove the `/* NEW * /`

packages/components/src/DeliveryReclamation.purs
```diff
@@ -0,0 +1,175 @@
...
+
+    claimNewDeliveryInput =
+      InputField.inputField
+        { type_: InputField.Radio
+        , placeholder: "Placeholder"
+        , name: "claim"
+        , onChange: \newClaim ->
+                      let
+                        updateClaim x = self.setState _ { claim = x }
+                        parsed = read <$> newClaim
```
vapaj: Instead of using `map` here, and having `parsed :: Maybe (Maybe User.DeliveryReclamationClaim)`, we could use bind to have `parsed :: Maybe User.DeliveryReclamationClaim`:
`parsed = read =<< newClaim`

packages/components/src/DeliveryReclamation.purs
```diff
@@ -0,0 +1,175 @@
...
+                      in
+                        case parsed of
```
vapaj: Here we can then actually call `setState` with `parsed` directly.

packages/components/src/DeliveryReclamation.purs
```diff
@@ -0,0 +1,175 @@
...
+  , onError   :: Persona.InvalidDateInput -> Effect Unit
+  }
+
+deliveryReclamation :: Props -> JSX
+deliveryReclamation = make component { initialState, render, didMount }
+
+initialState :: State
+initialState =
+  { publicationDate: Nothing
+  , claim: Nothing
```
vapaj: Should we have a default value here? 🤔

yarn.lock
```diff
@@ -826,8 +826,12 @@
   version "0.2.1"
   dependencies:
     basscss "^8.1.0"
+<<<<<<< HEAD
```
vapaj: What happened here? 😁

packages/components/src/DeliveryReclamation.purs
ae-mo: Yes, I started re-reading the Monads chapter only yesterday lol

packages/components/src/DeliveryReclamation.purs
```diff
@@ -0,0 +1,176 @@
...
+                        , className: "mt2 clearfix"
+                        }
+          }
+
+    publicationDayInput = dateInput self self.state.publicationDate "Utgivningsdatum"
+
+    claimExtensionInput =
+      InputField.inputField
+        { type_: InputField.Radio
+        , placeholder: "Placeholder"
```
vapaj: The `placeholder` prop kinda doesn't make sense when it's a radio button (maybe we should rethink `InputField` at some point), however let's have something meaningful here like `"extension"`.

packages/components/src/DeliveryReclamation.purs
```diff
@@ -0,0 +1,176 @@
...
+          liftEffect $ self.props.onLoading
+          User.createDeliveryReclamation self.props.userUuid self.props.subsno date'' claim'' >>=
+            case _ of
+              Right recl -> liftEffect $ self.props.onSuccess recl
+              Left invalidDateInput -> liftEffect $ self.props.onError invalidDateInput
+    submitForm _ Nothing = self.setState _ { validationError = Just "Välj ett alternativ." }
+    submitForm _ _ = Console.error "The entered information is incomplete."
+
+radioButtonOnChange :: Self ->  Maybe String -> Effect Unit
+radioButtonOnChange self newClaim = do
```
vapaj: We can drop the `do` here

less/Subscription.less
```diff
@@ -68,6 +68,16 @@
     cursor: pointer;
 }
 
+.subscription--delivery-reclamation-icon {
+    display: block;
+    width: 30px;
+    height: 30px;
+    background-position: center;
+    background-repeat: no-repeat, repeat;
```
vapaj: This is a funny rule 😅

Issue #259, opened 2020-02-19
Use google SignIn library in a civilized way


zudov:
Apparently there is a way to get user info without going through `Zi.Qt.U3`: [GoogleUser.getBasicProfile()](https://developers.google.com/identity/sign-in/web/reference#googleusergetbasicprofile). So far we were just barbarically inspecting the runtime representation of the response and read the alien-named fields directly.

PR #260, opened 2020-02-19
Quickfix for the change of internal GoogleUser structure

zudov:
Issue #259 for proper fix


packages/user/src/User/Login/Google/Google.js
```diff
@@ -16,6 +16,7 @@ function initSuccess(args) {
     GoogleAuth.isSignedIn.listen(setUser(args.onSuccess));
     if (GoogleAuth.isSignedIn.get()) {
       var user = GoogleAuth.currentUser.get();
+      console.info("GoogleAuth.currentUser", user)
```
vapaj: Maybe we don't need the logging here

PR #263, opened 2020-02-25
Enhance/add opinion type

lulzimfazlija:
show opinion type in article header details

### DIFF
--- a/apps/app-article/src/components/article-details.js
+++ b/apps/app-article/src/components/article-details.js
@@ -61,7 +61,7 @@ const ArticleDetails = (props) => {
                                 }
                                 {
                                     props.premium ?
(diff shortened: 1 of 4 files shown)


apps/app-article/src/components/article-details.js
```diff
@@ -61,7 +61,7 @@ const ArticleDetails = (props) => {
                                 }
                                 {
                                     props.premium ?
-                                        <PremiumBadge/>
+                                        <div className="article-info"><span className="article-opinion-type">{props.articleTypeDetails.title}</span> <PremiumBadge/></div>
```
vapaj: `articleTypeDetails` is not always returned from the API, so we need to check that it exists whenever we access its attributes.

PR #264, opened 2020-02-26
Make OpenApiClient module with utilities for using generated clients

zudov:
Just moving the copy-pasted code to a shared module and improve logging a bit.


vapaj:
I think you forgot to add the `OpenApiClient` module itself 😁

PR #269, opened 2020-03-03
Canceled orders

vapaj:
Catch canceled orders and show the `NewPurchase` view.

Also fix some warnings.

### DIFF
--- a/packages/components/src/Bottega.purs
+++ b/packages/components/src/Bottega.purs
@@ -33,13 +33,7 @@ readOrder orderObj = do
   orderStatus <- case read orderObj.status.state of
     Right status -> pure status
     Left err     -> pure UnknownState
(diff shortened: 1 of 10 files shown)


packages/components/src/Bottega.purs
```diff
@@ -3,44 +3,43 @@ module Bottega where
...
-               { state: orderStatus
-               , time: orderObj.status.time
-               }
+
+readOrder :: { number :: OrderNumber, user :: UUID, status :: { state :: Foreign, time :: String } } -> Aff Order
+readOrder orderObj = do
+  orderStatus <- case read orderObj.status.state of
+    Right status -> pure status
+    Left err     -> pure UnknownState
+  pure { number: orderObj.number
```
vapaj: This could be refactored a bit actually

PR #272, opened 2020-03-04
Vetrina packages from server

vapaj:
Fetch packages from server and for now, filter only `HblPremium` from there. If the package(s) wanted are not found or valid, show error page.


packages/components/src/Product/HblPremium.purs
```diff
@@ -0,0 +1,21 @@
...
+
+hblPremium :: Product.Product -> JSX
+hblPremium p = Product.product p
+
+toProduct :: Package -> Product.Product
+toProduct  { name, id, offers } =
+  { name: name
+  , id:   id
+  , description: "Alla artiklar på hbl.fi"
+  , price: maybe 100.0 (_.monthlyPrice >>> toNumber >>> (_ / 100.0)) $ Array.head offers
```
ae-mo: I don't get this, can you clarify?

vapaj: It actually doesn't make sense 😄 I forgot to fix it, good catch!

PR #278, opened 2020-03-09
Sentry tryout

vapaj:
Add sentry integration and send events from obvious places in Vetrina.

### DIFF
--- a/apps/vetrina/.env.development
+++ b/apps/vetrina/.env.development
@@ -1,2 +1,3 @@
 PERSONA_URL=https://persona.staging.ksfmedia.fi/v1
 BOTTEGA_URL=https://bottega.staging.ksfmedia.fi/v1
+SENTRY_DSN=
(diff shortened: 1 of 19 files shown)


apps/vetrina/.env.development
```diff
@@ -1,2 +1,3 @@
 PERSONA_URL=https://persona.staging.ksfmedia.fi/v1
 BOTTEGA_URL=https://bottega.staging.ksfmedia.fi/v1
+SENTRY_DSN=
```
ae-mo: I'm guessing this value depends on the local dev environment configuration?

vapaj: Yeah. Let's not commit the value here

ae-mo:
@vapaj can you point me to a place in your code where you actually send a `sentry` event? I have to confess I cannot find it 😆

PR #296, opened 2020-03-19
Ask to set password after purchase

vapaj:
If the user has not completed registration, ask them to set a password after purchase.


apps/vetrina/src/Vetrina/Main.purs
```diff
@@ -236,7 +236,17 @@ render self =
...
     ProcessPayment -> Spinner.loadingSpinner
     PurchaseFailed -> DOM.text "PURCHASE FAILED :~("
-    PurchaseDone -> PurchaseCompleted.completed { redirectArticleUrl: Nothing }
+    PurchaseDone ->
+      PurchaseCompleted.completed
+        -- TODO: The onError callback is invoked if setting the new password fails.
+        -- We should think how to handle this. Probably we don't want to
+        -- show an ORDER FAILED message, but rather just inform the user that
+        -- something went wrong and please try to set the password again some other time.
+        { onError: \_ -> pure unit
```
zudov: Should we log the error, into sentry or just console?

vapaj: Sentry, but I'm waiting for the layouts/copy for this anyway, so let's resolve these later.

PR #302, opened 2020-03-24
Fix cache control

vapaj:
Don't send empty cache-control header

### DIFF
--- a/packages/components/src/Persona.purs
+++ b/packages/components/src/Persona.purs
@@ -12,7 +12,7 @@ import Data.Generic.Rep (class Generic)
 import Data.Generic.Rep.Show (genericShow)
 import Data.JSDate (JSDate)
 import Data.List (fromFoldable)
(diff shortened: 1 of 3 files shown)


packages/components/src/Persona.purs
```diff
@@ -47,7 +47,7 @@ getUser invalidateCache uuid token =
   where
     headers =
       { authorization
-      , cacheControl: fromMaybe mempty maybeCacheControl
+      , cacheControl: toNullable maybeCacheControl
```
ae-mo: So this way the `cache-control` header doesn't appear?

Btw wasn't this introduced in a recent PR I reviewed?

ae-mo: Oh no that was in `vetrina`

vapaj: It was yeah, it's a shared module though 😄

vapaj: Yes, if we set the `cacheControl` to `null`, it won't appear in the request

Issue #317, opened 2020-04-02
Vetrina layouts: landing screen


vapaj:
Implement first screen of vetrina according to layout. The red square marks the things that should be included in vetrina. Note that we cannot implement the checkbox about sending offers by email yet, so let's leave it out from this issue.

![vetrina-view-1](https://user-images.githubusercontent.com/6303323/78237754-6a1ef280-74e4-11ea-935a-58cab1ee71b0.png)

PR #318, opened 2020-04-02
Vetrina layouts: landing screen

vapaj:
Add #317

Also changed the functionality so that `vetrina` will get the products wished to show through `Props`, not by fetching the things itself.

Also `label` is an optional field of `InputField` now.


apps/vetrina/src/Vetrina/Main.purs
```diff
@@ -42,7 +38,7 @@ import Vetrina.Purchase.Completed as PurchaseCompleted
 
 foreign import sentryDsn_ :: Effect String
 
-type Props = { onClose :: Effect Unit }
+type Props = { onClose :: Effect Unit, products :: Array Product }
```
ae-mo, line 41: Why this formatting?

vapaj: Well, first of all it makes sense that whoever includes vetrina, also somehow tells it which products to show. Also, we don't want to hard code the descriptions of the products to vetrina, so it makes sense to just take the full products in props.

ae-mo: I just meant why the object's items are on the same line 😅

vapaj: Oh 🤠 It's such a short definition it fits nicely in one line

ae-mo: But apparently it's common in the code, so I won't be bothering anymore 😃

vapaj: Yeah, I don't know... I mean
```
type Props =
  { onClose :: Effect Unit }
```
would look a bit funny (although I suspect we have this style in the code as well :D)

But you're right, if there's > 1 attributes, they should be formatted in separate lines:
```
type Props =
  { onClose :: Effect Unit
  , products :: Array Product
  }
```

PR #320, opened 2020-04-03
show styles of differenc components based on paper type, improve layo…

lulzimfazlija:
change factbox to generic box


apps/app-article/src/components/article-content.js
```diff
@@ -17,8 +18,8 @@ class Content extends Component {
                 return this.renderImage(block, key);
             } else if (blockType === 'html') {
                 return this.renderHtml(block, key);
-            } else if (blockType === 'factbox') {
```
vapaj, line 20: Let's actually not remove this entirely, as it will break cached articles for instance. We could modify the object a bit, and pass it to `renderGenericBox`.

apps/app-article/src/components/article-content.js
```diff
@@ -70,14 +71,14 @@ class Content extends Component {
...
     }
 
-    renderFactBox(block, key) {
+    renderGenericBox(block, key) {
         return (
-            <div className={"factBox"} key={key} id={"factBox-" + key}>
-                <div className={"fakta"}>FAKTA</div>
-                <h4>{block.factBox.title}</h4>
+            <div className={`genericBox genericBox-border-${getBrandValueParam()}`} key={key} id={"genericBox-" + key}>
+                <div className={"fakta"}>{block.box && block.box.headline}</div>
```
vapaj, line 77: Two things:

Let's rename the class `"fakta"` to `"genericBox-headline"` or something.

If `box.type === "fact"`, we should show a default headline `"FAKTA"`.

apps/app-article/src/components/article-content.js
```diff
@@ -18,7 +19,13 @@ class Content extends Component {
             } else if (blockType === 'html') {
                 return this.renderHtml(block, key);
             } else if (blockType === 'factbox') {
-                return this.renderFactBox(block, key);
+                block.box.headline = 'Facta';
```
vapaj, line 22: Should be "FAKTA"

vapaj: Also, can you do this? If `box` does not exist in `block`,  this would not work.

vapaj: Could we just pass an object of
```
{
  box:
  {
    headline: "FAKTA",
    title: block.factBox.title,
    ... 
   }
}
```
to `renderGenericBox` ?

apps/app-article/src/components/article-content.js
```diff
@@ -70,14 +82,14 @@ class Content extends Component {
...
     }
 
-    renderFactBox(block, key) {
+    renderGenericBox(block, key) {
         return (
-            <div className={"factBox"} key={key} id={"factBox-" + key}>
-                <div className={"fakta"}>FAKTA</div>
-                <h4>{block.factBox.title}</h4>
+            <div className={`genericBox genericBox-border-${getBrandValueParam()}`} key={key} id={"genericBox-" + key}>
+                <div className={"genericBox-headline"}>{block.box && block.box.type === "fact" ? 'FAKTA': ''}</div>
```
vapaj, line 88: So the logic should be
```
if headline
then headline
else if type == fact
     then "FAKTA"
     else ""
```

lulzimfazlija: @vapaj if the current last commit does not fix it, should smth like this be ok:

```
 {
     block.box && block.box.headline &&  
           <div className={"genericBox-headline"}>{block.box.headline}</div>
 }
 {
     block.box && block.box.type &&  
           <div className={"genericBox-headline"}>{block.box.type === "fact" ? 'FAKTA': ''}</div>
 }
```

PR #333, opened 2020-04-14
Fix/subscription failure

### LINKED ISSUE #299: Handle SubscriptionExists
While we are checking if the customer already has the subscription in Vetrina, the client side and server side might not be in sync. So handle `SubscriptionExists` errors we get from Bottega.

ae-mo:
Related to #299.

### DIFF
--- a/apps/vetrina/src/Vetrina/Main.purs
+++ b/apps/vetrina/src/Vetrina/Main.purs
@@ -193,7 +193,7 @@ pollOrder setState state@{ logger } (Right order) = do
         chooseAccountStatus user
           | user.hasCompletedRegistration = ExistingAccount
           | otherwise = NewAccount
-    OrderFailed    -> liftEffect do
+    OrderFailed reason -> liftEffect do
       logger.error $ Error.orderError "Order failed for customer"
       setState _ { purchaseState = PurchaseFailed }
     OrderCanceled  -> liftEffect do
--- a/packages/components/src/Bottega.purs
+++ b/packages/components/src/Bottega.purs
@@ -32,18 +32,17 @@ getOrder { userId, authToken } orderNumber = do
     authorization = oauthToken authToken
     authUser = unsafeToForeign userId
 
-readOrder :: { number :: OrderNumber, user :: UUID, status :: { state :: Foreign, time :: String, failReason :: Nullable Foreign } } -> Aff Order
+readOrder :: { number :: OrderNumber, user :: UUID, status :: { state :: Foreign, time :: String, failReason :: Nullable String } } -> Aff Order
 readOrder orderObj = do
   orderStatus <- case read orderObj.status.state of
     Right status -> pure status
     Left err     -> pure UnknownState
   failureReason <- case orderStatus of
     UnknownState -> pure Nothing
-    _ -> case toMaybe orderObj.status.failReason of
-      Just reason -> case read reason of
-                       Right result -> pure (Just result)
-                       Left err     -> pure Nothing
-      Nothing     -> pure Nothing
+    _ -> pure $ do
+      reasonString <- toMaybe orderObj.status.failReason
+      reason       <- parseFailReason reasonString
+      pure reason
   pure $ orderObj { status { state = orderStatus, failReason = failureReason } }
 
 payOrder :: UserAuth -> OrderNumber -> PaymentMethod -> Aff PaymentTerminalUrl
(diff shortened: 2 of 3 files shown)


packages/components/src/Bottega.purs
```diff
@@ -30,12 +32,19 @@ getOrder { userId, authToken } orderNumber = do
     authorization = oauthToken authToken
     authUser = unsafeToForeign userId
 
-readOrder :: { number :: OrderNumber, user :: UUID, status :: { state :: Foreign, time :: String } } -> Aff Order
+readOrder :: { number :: OrderNumber, user :: UUID, status :: { state :: Foreign, time :: String, failReason :: Nullable Foreign } } -> Aff Order
```
vapaj, line 35: I think `failReason` can just be `Nullable String`

ae-mo: But then how do I parse an object of type `OrderStatusFailReason` out of it?

vapaj: Hmm, `state` seems to be `Foreign` as well 🤔 Should be `String` too.

vapaj: Basically you are already doing it here
```
    orderStatusStateString :: String <- readImpl foreignOrderStatusState
    case orderStatusStateString of
      "created"   -> pure OrderCreated
      "started"   -> pure OrderStarted
      "completed" -> pure OrderCompleted
      "failed"    -> pure OrderFailed
      "canceled"  -> pure OrderCanceled
      _           -> pure UnknownState
```

ae-mo: But wait, the signature of the `readImpl` function is `readImpl :: Foreign -> F a`: 
https://pursuit.purescript.org/packages/purescript-simple-json/7.0.0/docs/Simple.JSON#v:readImpl

So isn't `Nullable Foreign` correct?

packages/components/src/Bottega.purs
```diff
@@ -50,14 +59,15 @@ getPackages = callApi packagesApi "packageGet" [] {}
...
+  , user       :: UUID
+  , status     :: OrderStatus
   }
 
 type OrderStatus =
-  { state  :: OrderStatusState
-  , time   :: String
+  { state      :: OrderStatusState
+  , time       :: String
+  , failReason :: Maybe OrderStatusFailReason
```
vapaj, line 70: Actually, because we know `failReason` only exists when the state is failed, we could just add it to the `OrderFailed` constructor (instead of having it in the `OrderStatus` record):
```
data OrderStatusState
  = OrderCreated
  | OrderStarted
  | OrderCompleted
  | OrderFailed OrderStatusFailReason
  | OrderCanceled
  | UnknownState
```

packages/components/src/Bottega.purs
```diff
@@ -30,12 +32,18 @@ getOrder { userId, authToken } orderNumber = do
...
     Right status -> pure status
     Left err     -> pure UnknownState
-  pure $ orderObj { status { state = orderStatus } }
+  failureReason <- case orderStatus of
+    UnknownState -> pure Nothing
+    _ -> pure $ do
+      reasonString <- toMaybe orderObj.status.failReason
+      reason       <- parseFailReason reasonString
+      pure reason
+  pure $ orderObj { status { state = orderStatus, failReason = failureReason } }
```
ae-mo, line 46: @vapaj in this version `orderObj.status.failReason` is a `Nullable String`.

The function `parseFailReason` has signature `String -> Maybe OrderStatusFailReason`.

Is this what you meant? Can I refactor the `status` field in a similar fashion (except it's not `Nullable`)?

vapaj: > The function `parseFailReason` has signature `String -> Maybe OrderStatusFailReason`

I think it does not have to be a `Maybe`, as we have the `UnknownReason` constructor in case we cannot parse it (so `String -> OrderStatusFailReason`).

> Can I refactor the `status` field in a similar fashion (except it's not `Nullable`)

Yep 👍

ae-mo: > I think it does not have to be a Maybe

That was done like this so I could use monadic style to propagate the possible `Nothing` coming from the `toMaybe` call without explicitly handling it

vapaj: Because `Maybe` implements `Functor`, you can use `map` here to achieve this:
```
reason <- map parseFailReason $ toMaybe orderObj.status.failReason
```
(https://pursuit.purescript.org/packages/purescript-maybe/4.0.1/docs/Data.Maybe#t:Maybe)

apps/vetrina/src/Vetrina/Main.purs
```diff
@@ -193,7 +193,7 @@ pollOrder setState state@{ logger } (Right order) = do
         chooseAccountStatus user
           | user.hasCompletedRegistration = ExistingAccount
           | otherwise = NewAccount
-    OrderFailed    -> liftEffect do
+    OrderFailed reason -> liftEffect do
```
vapaj, line 196: Now we need to handle the `SubscriptionExistsError` case

ae-mo: How could we do it?

In #299 you said we should just close `vetrina` if I'm correct, but doesn't this lack of transparency to the user? If she made an order, maybe she would like to know what happend to it

PR #338, opened 2020-04-19
genericbox show show more icon only when there is more than 400 chara…

lulzimfazlija:


apps/app-article/src/components/article-content.js
```diff
@@ -82,9 +82,22 @@ class Content extends Component {
         );
     }
 
+    countBlockChatacters(block){ 
+        if(block){
+            const totalCharacters = block.reduce((acc, item, currentIndex) => {
+                if(currentIndex === 0){
```
vapaj, line 88: What is the idea here? This will never calculate the length of the first string, meaning, if I pass `countBlockChatacters` an array of one string, the count will be 0.

lulzimfazlija: that is true, I made it so keeping in mind that we will have always this review section on first element which contains html code, but in fact for factbox it is not the case.

PR #341, opened 2020-04-27
Fix/vetrina error view

### LINKED ISSUE #336: Error view
Add an error view according to:
image (1)

However, the correct text should be

```
Något gick fel med ditt inköp!
Klicka nedan för pröva på nytt eller ta kontakt med vår kundservice på pren@hbl.fi.
Vi beklagar strulet!
Pröva på nytt
```

ae-mo:
Closes #336.

### DIFF
--- a/apps/vetrina/src/Vetrina/Main.purs
+++ b/apps/vetrina/src/Vetrina/Main.purs
@@ -241,19 +241,20 @@ render self =
         }
   where
     onRetry = do
-      Aff.launchAff_ $ tryMagicLogin self
-      self.setState _ { purchaseState = NewPurchase }
+      case self.state.user of
+        Just user -> self.setState _ { purchaseState = NewPurchase, accountStatus = LoggedInAccount user }
+        Nothing   -> self.setState _ { purchaseState = NewPurchase }
 
 vetrinaContainer :: Self -> Array JSX -> JSX
 vetrinaContainer self@{ state: { purchaseState } } children =
   let errorClassString = "vetrina--purchase-error"
       errorClass       = case purchaseState of
-                           PurchaseFailed          -> " " <> errorClassString
-                           PurchaseUnexpectedError -> " " <> errorClassString
+                           PurchaseFailed          -> errorClassString
+                           PurchaseUnexpectedError -> errorClassString
                            otherwise               -> mempty
   in
     DOM.div
-      { className: "vetrina--container" <> errorClass
+      { className: "vetrina--container " <> errorClass
       , children
       }
 
--- a/apps/vetrina/src/Vetrina/Main.purs
+++ b/apps/vetrina/src/Vetrina/Main.purs
@@ -178,7 +178,9 @@ pollOrder setState state@{ logger } (Right order) = do
       setState _ { purchaseState = PurchaseFailed }
     OrderCanceled  -> liftEffect do
       logger.log "Customer canceled order" Sentry.Info
-      setState _ { purchaseState = NewPurchase }
+      case state.user of
+        Just user -> setState _ { purchaseState = NewPurchase, accountStatus = LoggedInAccount user }
+        Nothing   -> setState _ { purchaseState = NewPurchase }
     OrderCreated   -> pollOrder setState state =<< User.getOrder order.number
     UnknownState   -> liftEffect do
       logger.error $ Error.orderError "Got UnknownState from server"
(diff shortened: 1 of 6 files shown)


ae-mo:
I would like to discuss what should happen when the user hits the retry button.

Initially I thought I could simply reset the state to the initial state, but that is causing problems and I think the reason is that if I already tried to authenticate, my credentials are now saved in local storage.

What I'm experiencing is that `vetrina` hangs, showing the loader indefinitely after I hit the retry button.

vapaj:
> I think the reason is that if I already tried to authenticate, my credentials are now saved in local storage.

Why is this a problem?

apps/vetrina/src/Vetrina/Main.purs
```diff
@@ -231,7 +239,10 @@ render self =
             , children: [ DOM.text "OK" ]
             }
         ]
-    PurchaseUnexpectedError -> DOM.text "SOMETHING WENT HORRIBLY WRONG SERVER SIDE"
+    PurchaseUnexpectedError -> vetrinaContainer $ Array.singleton $
+      Purchase.Error.error
+        { onRetry: self.setState \_ -> initialState
```
vapaj, line 244: You cannot simply set it to `initialState`, as for instance the loading spinner is then switched on (`isLoading: Just Spinner.Loading)`. Also it's wiping all the state we actually need (sentry logger, user, product selection).

Would it be enough to set the `purchaseState` to `NewPurchase`?

vapaj: This is what we're doing when the user cancels the process (from NETS terminal) https://github.com/KSF-Media/affresco/blob/master/apps/vetrina/src/Vetrina/Main.purs#L178-L180, and the action should be same here.

Although, I noticed this is not enough, as we are not recognizing the user we just created. The view should be the new-purchase-with-logged-in-user one.

vapaj: We should also set the `accountStatus` depending on what the `state.user` is, like here https://github.com/KSF-Media/affresco/blob/master/apps/vetrina/src/Vetrina/Main.purs#L124-L129

ae-mo: It seems to be working, but what about the `newOrder`, `OrderFailure`, `PaymentMethod`? Don't we need to wipe them if we are starting anew?

vapaj: `paymentMethod` is always `CreditCard` for now, so that's fine. However, I was looking at `newOrder` and `orderFailure` of `state`, and thinking do we even use those anywhere (except for _setting_ them)? 😄 If not, just remove them from `state`.

apps/vetrina/src/Vetrina/Main.purs
```diff
@@ -231,14 +235,27 @@ render self =
...
-    { className: "vetrina--container"
-    , children
-    }
+    PurchaseUnexpectedError -> vetrinaContainer self $ Array.singleton $
+      Purchase.Error.error
+        { onRetry: onRetry
+        }
+  where
+    onRetry = do
+      Aff.launchAff_ $ tryMagicLogin self
```
vapaj, line 244: We don't need to actually login, because we already have the user in the state (and if we don't, the magic login would not help, as in that case we don't have the user credentials in local storage). All we need to do, is just set the `purchaseState` to `NewPurchase` and `accountStatus` according to whatever the `state.user` is.

So something like:
```purescript
case state.user of
  Just user -> self.state { purchaseState = NewPurchase, accountStatus = LoggedInAccount user }
  Nothing   -> self.state { purchaseState = NewPurchase }
```

apps/vetrina/src/Vetrina/Main.purs
```diff
@@ -231,14 +235,27 @@ render self =
...
+      self.setState _ { purchaseState = NewPurchase }
+
+vetrinaContainer :: Self -> Array JSX -> JSX
+vetrinaContainer self@{ state: { purchaseState } } children =
+  let errorClassString = "vetrina--purchase-error"
+      errorClass       = case purchaseState of
+                           PurchaseFailed          -> " " <> errorClassString
```
vapaj, line 251: Just a personal opinion, but somehow just adding a space to the end of `className: "vetrina--container " ...` seems cleaner to me, rather than adding the space here 😄

PR #346, opened 2020-05-05
show most read and latest article based on paper type

lulzimfazlija:


apps/app-article/src/article-service.js
```diff
@@ -11,13 +11,13 @@ const articleApi = {
 
     },
     getLatestArticles() {
-        return fetch(config.apiUrl + "latest?start=0&limit=10", {
+        return fetch(config.apiUrl + "latest?start=0&limit=10&paper=" + getBrandValueParam() + "", {
```
vapaj, line 14: Why the empty string at the end? 🤔

lulzimfazlija: Thats a good question, will fix no need for it 😀

PR #347, opened 2020-05-05
Add support for moving DIGIHEL ad into text body

jockeF:
The DIGIHEL banner is very low on the page but is sold as a in text banner. As there is no easy way to output a banner slot while rolling out the article text this script looks for a spot in the midsection of the text where an banner ad can be placed. If a good spot is found it moves the banner there.


scripts/app/adMover.js
```diff
@@ -0,0 +1,73 @@
...
+        let response = false;
+        expectedType = expectedType.toUpperCase(); // nodeName is always uppercase
+        // check that the surrounding paragraphs are OK when placing the in text ad
+        let pBefore = ptag.previousElementSibling;
+        let pBeforeContent = pBefore.innerHTML.replace(/[^A-Za-z0-9]/g, '');
+        let pBeforeClass = pBefore.classList.contains(expectedClass);
+        let pSelf = ptag;
+        let pSelfContent = pSelf.innerHTML.replace(/[^A-Za-z0-9]/g, '');
+        let pSelfClass = pSelf.classList.contains(expectedClass);
+        if ((pBeforeContent !== "" && pSelfContent !== "" && !(/^\s*$/.test(pBeforeContent)) && !(/^\s*$/.test(pSelfContent))) && (pSelf.nodeName === expectedType && pBefore.nodeName === expectedType) && (pBeforeClass && pSelfClass)) {
```
vapaj, line 14: I think the `!(/^\s*$/.test(pBeforeContent))` will also cover the `pBeforeContent !== ""` case, so the latter can be removed.

jockeF: You are right. I do think that in the case of an empty string the first test i a lot more performant. But in practice that is probably not significant as this at worst will be run like two times. Also that line is quite ugly so cleaning it up a bit is a value in itself.

ae-mo:
I guess this can be merged?

PR #358, opened 2020-05-11
Fix/vetrina login view

### LINKED ISSUE #342: Handle invalid credentials
Currently we are rendering the error view when user inserts invalid credentials.

We should rather show similar error message what we have in the `User` component:
Screenshot_2020-04-28 Mitt Konto

ae-mo:
Closes #342.

Contains a refactoring of `Vetrina` module too.

### DIFF
--- a/packages/vetrina/src/Vetrina/Purchase/NewPurchase.purs
+++ b/packages/vetrina/src/Vetrina/Purchase/NewPurchase.purs
@@ -5,6 +5,7 @@ import Prelude
 import Control.Alt ((<|>))
 import Data.Array (all, cons, intercalate)
 import Data.Array as Array
+import Data.Foldable (foldMap)
 import Data.Either (Either(..))
 import Data.Maybe (Maybe(..), fromMaybe)
 import Data.Nullable (toMaybe)
@@ -42,7 +43,7 @@ type State =
 type Props =
   { accountStatus                 :: AccountStatus
   , products                      :: Array Product
-  , errorMessage                  :: JSX
+  , errorMessage                  :: Maybe String
   , mkPurchaseWithNewAccount      :: NewAccountForm -> Effect Unit
   , mkPurchaseWithExistingAccount :: ExistingAccountForm -> Effect Unit
   , mkPurchaseWithLoggedInAccount :: User -> { | PurchaseParameters } -> Effect Unit
@@ -158,7 +159,7 @@ form self = DOM.form $
     -- NOTE: We need to have `emailInput` here (opposed to in `children`),
     -- as we don't want to re-render it when `accountStatus` changes.
     -- This will keep cursor focus in the input field.
-  , children: self.props.errorMessage `cons` (emailInput self self.state.accountStatus `cons` children)
+  , children: (foldMap formatErrorMessage self.props.errorMessage) `cons` (emailInput self self.state.accountStatus `cons` children)
   }
   where
     onSubmit = handler preventDefault $ case self.state.accountStatus of
@@ -264,6 +265,9 @@ isFormInvalid validations
   | Left errs <- toEither validations
   = not $ all isNotInitialized errs
   | otherwise = false
+
+formatErrorMessage :: String -> JSX
+formatErrorMessage message = InputField.errorMessage message
 
 emailInput :: Self -> AccountStatus -> JSX
 emailInput _ (LoggedInAccount _) = mempty
--- a/packages/vetrina/src/Vetrina/Purchase/NewPurchase.purs
+++ b/packages/vetrina/src/Vetrina/Purchase/NewPurchase.purs
@@ -40,14 +40,15 @@ type State =
   }
 
 type Props =
-  { accountStatus :: AccountStatus
-  , products :: Array Product
-  , mkPurchaseWithNewAccount :: NewAccountForm -> Effect Unit
+  { accountStatus                 :: AccountStatus
+  , products                      :: Array Product
+  , errorMessage                  :: JSX
+  , mkPurchaseWithNewAccount      :: NewAccountForm -> Effect Unit
   , mkPurchaseWithExistingAccount :: ExistingAccountForm -> Effect Unit
   , mkPurchaseWithLoggedInAccount :: User -> { | PurchaseParameters } -> Effect Unit
-  , paymentMethod :: PaymentMethod
-  , productSelection :: Maybe Product
-  , onLogin :: Effect Unit
+  , paymentMethod                 :: PaymentMethod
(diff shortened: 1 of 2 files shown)


packages/vetrina/src/Vetrina/Purchase/NewPurchase.purs
```diff
@@ -40,14 +40,15 @@ type State =
   }
 
 type Props =
-  { accountStatus :: AccountStatus
-  , products :: Array Product
-  , mkPurchaseWithNewAccount :: NewAccountForm -> Effect Unit
+  { accountStatus                 :: AccountStatus
+  , products                      :: Array Product
+  , errorMessage                  :: JSX
```
vapaj, line 45: Let's change this to `String` and let `NewPurchase` decide what `JSX` to render.

packages/vetrina/src/Vetrina/Vetrina.purs
```diff
@@ -85,11 +85,9 @@ data PurchaseState
   = NewPurchase
   | CapturePayment PaymentTerminalUrl
   | ProcessPayment
-  | PurchaseFailed
+  | PurchaseFailed (Maybe OrderFailure)
```
vapaj, line 88: This should be `PurchaseFailed OrderFailure`, as we should always know what error happened.

packages/vetrina/src/Vetrina/Vetrina.purs
```diff
@@ -98,6 +96,7 @@ data OrderFailure
   | AuthenticationError
   | ServerError
   | UnrecognizedError String
+  | UnexpectedError
```
vapaj, line 99: Let's either have `UnrecognizedError String` or `UnexpectedError`, but not both.

packages/vetrina/src/Vetrina/Vetrina.purs
```diff
@@ -166,10 +165,10 @@ didUpdate self _ = Aff.launchAff_ $ stopOrderPollerOnCompletedState self
 stopOrderPollerOnCompletedState :: Self -> Aff Unit
 stopOrderPollerOnCompletedState self =
   case self.state.purchaseState of
-    PurchaseFailed      -> killOrderPoller self.state
-    PurchaseCompleted _ -> killOrderPoller self.state
-    NewPurchase         -> killOrderPoller self.state
-    _                   -> pure unit
+    PurchaseFailed Nothing -> killOrderPoller self.state
```
vapaj, line 168: We need to stop the order polling if any `PurchaseFailed` happens.

packages/vetrina/src/Vetrina/Purchase/NewPurchase.purs
ae-mo: But this way `NewPurchase` needs to know what kind of order failure was there correct? At least if I understood what you mean by let `NewPurchase` decide what `JSX` to render.

packages/vetrina/src/Vetrina/Vetrina.purs
ae-mo: What error would you return here?

https://github.com/KSF-Media/affresco/blob/master/packages/vetrina/src/Vetrina/Vetrina.purs#L205

packages/vetrina/src/Vetrina/Vetrina.purs
ae-mo: Semantically they are slightly different, what would be the criterion to decide? Maybe what's the more general? In that case I would go with `UnexpectedError`.

packages/vetrina/src/Vetrina/Purchase/NewPurchase.purs
vapaj: If we are always showing the error above the input field, it's enough to know what the message is, right?

So the thing in `Vetrina` becomes:
```purescript
orderErrorMessage :: OrderFailure -> String
orderErrorMessage failure =
  case failure of
    AuthenticationError -> "Kombinationen av e-postadress och lösenord finns inte"
    _ -> DOM.text "Något gick fel. Vänligen försök om en stund igen."
```

packages/vetrina/src/Vetrina/Vetrina.purs
vapaj: Yeah, maybe with the message as well: `UnexpectedError String`. What do you think?

packages/vetrina/src/Vetrina/Vetrina.purs
vapaj: Depends on what the `OrderStatusFailReason` is (https://github.com/KSF-Media/affresco/blob/master/packages/components/src/Bottega.purs#L82-L89).

I'm currently implementing the `SubscriptionExistsError` one, but I think in other cases we can default to `ServerError` or `UnexpectedError` even though they're not exactly true (It probably would not be beneficial to reflect all the `OrderStatusFailReason` values to the user interface)

packages/vetrina/src/Vetrina/Vetrina.purs
ae-mo: I agree 👍

packages/vetrina/src/Vetrina/Vetrina.purs
```diff
@@ -238,10 +236,33 @@ render self = vetrinaContainer self $
...
     CapturePayment url -> netsTerminalIframe url
     ProcessPayment -> Spinner.loadingSpinner
-    PurchaseFailed ->
-      Purchase.Error.error
-        { onRetry
-        }
+    PurchaseFailed failure ->
+      case failure of
+        SubscriptionExists ->
+          DOM.div_
```
vapaj, line 242: I guess this got lost in the rebase conflict mess 😄 Should be 
```purescript
Purchase.SubscriptionExists.subscriptionExists
  { onClose: self.props.onClose }
```

packages/vetrina/src/Vetrina/Vetrina.purs
```diff
@@ -333,14 +347,14 @@ mkPurchase self@{ state: { logger } } validForm affUser = Aff.launchAff_ $ Spinn
         startOrderPoller self.setState newState order
     Left err -> do
       case err of
-        UnrecognizedError e -> liftEffect $ logger.error $ Error.orderError $ "Failed to place an order: " <> e
+        UnexpectedError e -> liftEffect $ logger.error $ Error.orderError $ "Failed to place an order: " <> e
         _ -> pure unit
 
       let errState = { user: hush eitherUser } `merge` case err of
```
vapaj, line 353: I think we lost the case for `AuthenticationError` here

PR #360, opened 2020-05-12
Unpause subscription

zudov:


less/Subscription.less
```diff
@@ -73,6 +73,27 @@
...
+        width: 30px;
+        height: 30px;
+        text-align: center;
+        font-size: 30px;
+        font-weight: lighter;
+        opacity: 0.7;
+        content: '✕';
+    }
+}
+
```
ae-mo, line 96: @vapaj the icon is basically composed of the one use for pause subscription with an 'x' over it. Maybe we could get on from Steffen later on?

vapaj: Yeah, ask Steffen about it, he might have ideas.

apps/vetrina/.build-size-sha256
```diff
@@ -0,0 +1,3 @@
+# SHA256 of the content of the build directory.
```
vapaj, line 1: No need to add this file

packages/components/src/Persona.purs
```diff
@@ -102,6 +102,20 @@ pauseSubscription uuid subsno startDate endDate token = do
...
 
+unpauseSubscription :: UUID -> Int -> Maybe DateTime -> Token -> Aff Subscription
+unpauseSubscription uuid subsno maybeStartDate token = do
+  let maybeStartDateISO = formatDate <$> maybeStartDate
+  callApi usersApi "usersUuidSubscriptionsSubsnoUnpausePost"
+    ([ unsafeToForeign uuid
+     , unsafeToForeign subsno
+     ] <> case maybeStartDateISO of
+            Nothing -> []
+            Just startDateISO -> [ unsafeToForeign { startDate: startDateISO } ]
```
vapaj, line 113: Why the `startDate` here though? Didn't we remove it from persona anyway? 🤔

ae-mo: Right, I will remove it from here too

packages/components/src/Subscription/Component.purs
```diff
@@ -160,7 +163,11 @@ render self@{ props: props@{ subscription: sub@{ package } } } =
         where
           asyncWrapper = AsyncWrapper.asyncWrapper
             { wrapperState: self.state.wrapperProgress
-            , readyView: actionsContainer [ pauseIcon, temporaryAddressChangeIcon, deliveryReclamationIcon ]
+            , readyView: actionsContainer $ [ pauseIcon ]
```
vapaj, line 166: We don't need to wrap these in arrays, I think it would be cleaner/easier to read as
```purescript
[ pauseIcon
, if maybe true Array.null self.state.pausedSubscriptions
  then mempty
  else removeSubscriptionPauses
, temporaryAddressChangeIcon
, deliveryReclamationIcon
]
```

PR #364, opened 2020-05-13
bigger ad marker

jockeF:
A bigger headline and footer for ads to make them more distinct from editorial content.


scripts/app/gamAppAds.js
```diff
@@ -182,7 +184,11 @@ var ksfDfp = {};
             divToClose.style.display = "none";
             var headerToClose = document.getElementsByClassName('ksfDFPadHeader ' + slot);
             if (headerToClose[0] !== "undefined") {
```
vapaj, line 186: This will probably be always true. Should it be `headerToClose[0] !== undefined`?

jockeF: Good catch. I updated it to use typeof and also check for null which is the most likely problem in this case.

PR #367, opened 2020-05-13
Fix/vetrina error messages

### LINKED ISSUE #365: Hide error message upon email editing
In `vetrina`, we should hide any error message being displayed above the login form as soon as the user edits their email address.

ae-mo:
Closes #365.

### DIFF
--- a/packages/vetrina/src/Vetrina/Purchase/NewPurchase.purs
+++ b/packages/vetrina/src/Vetrina/Purchase/NewPurchase.purs
@@ -37,7 +37,7 @@ type State =
        }
   , accountStatus       :: AccountStatus
   , serverErrors        :: Array (Form.ValidationError FormInputField)
-  , displayErrorMessage :: Boolean
+  , errorMessage        :: JSX
   , productSelection    :: Maybe Product
   , paymentMethod       :: Maybe PaymentMethod
   }
@@ -100,7 +100,7 @@ newPurchase props = make component
                       }
                   , accountStatus: NewAccount
                   , serverErrors: []
-                  , displayErrorMessage: isJust props.errorMessage
+                  , errorMessage: foldMap formatErrorMessage props.errorMessage
                   , productSelection: Nothing
                   , paymentMethod: Nothing
                   }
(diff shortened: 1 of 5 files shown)


packages/vetrina/src/Vetrina/Purchase/NewPurchase.purs
```diff
@@ -36,6 +37,7 @@ type State =
        }
   , accountStatus       :: AccountStatus
   , serverErrors        :: Array (Form.ValidationError FormInputField)
+  , displayErrorMessage :: Boolean
```
vapaj, line 40: We kind of have the "to show or not to show the error message" logic already in the `errorMessage :: Maybe String` type we get from props. So instead of having this boolean, we could just have `errorMessage :: Maybe String` (or perhaps even `Maybe JSX`) in the state, and change the calls to `self.props.errorMessage` to `self.state.errorMessage`.

ae-mo: To this argument I counter that we already have the the message in the props, but either is fine for me

ae-mo: And thinking more about it, the `errorMessage :: Maybe String` in the only tells us whether, when we instantiate the component from the `vetrina` parent component, we need to display the message. But if we need to hide it later is not determined by it.

But I'm doing as you say and this does simplify the code

vapaj: I think it's more descriptive to deal straight with Maybes if possible, and avoid adding Booleans to the mix. In my opinion the Boolean does not bring much to the table here. If we want to make changes to the `errorMessage` it's perfectly fine to have it in the state.

If we use Maybe, it's pretty straight forward as we have now:
```purescript
if (isJust errorMessage)
then renderErrorMessage errorMessage
else mempty
```
But for instance, if we rely on a combination of Maybe and Boolean, we can end up in a situation of `errorMessage = Nothing` and `displayErrorMessage = true` which would not make sense.

vapaj: > the props only tells us whether, when we instantiate the component from the vetrina parent component, we need to display the message. But if we need to hide it later is not determined by it.

Exactly

PR #373, opened 2020-05-14
Add external JSX rendering to Vetrina component

### LINKED ISSUE #352: Unexpected server error view
Add view for an unexpected non-user-related error that we don't know how to recover from, like not getting a proper `Product` as a prop.

ae-mo:
Closes #352.

### DIFF
--- a/packages/vetrina/src/Vetrina/Vetrina.purs
+++ b/packages/vetrina/src/Vetrina/Vetrina.purs
@@ -41,12 +41,14 @@ type JSProps =
   { onClose :: Nullable (Effect Unit)
   , onLogin :: Nullable (Effect Unit)
   , products :: Nullable (Array JSProduct)
+  , unexpectedError :: Nullable JSX
   }
 
 type Props =
-  { onClose  :: Effect Unit
-  , onLogin  :: Effect Unit
-  , products :: Either Error (Array Product)
+  { onClose         :: Effect Unit
+  , onLogin         :: Effect Unit
+  , products        :: Either Error (Array Product)
+  , unexpectedError :: Maybe JSX
   }
 
 fromJSProps :: JSProps -> Props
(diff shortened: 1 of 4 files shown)


packages/vetrina/src/Vetrina/Vetrina.purs
```diff
@@ -41,12 +41,14 @@ type JSProps =
...
   }
 
 type Props =
-  { onClose  :: Effect Unit
-  , onLogin  :: Effect Unit
-  , products :: Either Error (Array Product)
+  { onClose         :: Effect Unit
+  , onLogin         :: Effect Unit
+  , products        :: Either Error (Array Product)
+  , unexpectedError :: Maybe JSX
```
vapaj, line 51: I think we should always expect to have this as prop (not `Maybe`), because there might be unexpected errors happening even when the parent isn't aware of them.

packages/vetrina/src/Vetrina/Vetrina.purs
```diff
@@ -121,34 +126,41 @@ initialState =
...
+          -- When user has been fetched, hide loading spinner
+          (liftEffect $ self.setState \s -> s { isLoading = Nothing })
+          do
+            -- Try to login with local storage information and set user to state
+            tryMagicLogin self
+
+            products <- liftEffect $ case self.props.products of
+              Right p -> pure p
+              Left err -> do
+                self.setState _ { purchaseState = PurchaseFailed $ UnexpectedError ""  }
```
vapaj, line 156: For instance this would not render anything, although we cannot render vetrina successfully, as the products were broken.

packages/vetrina/src/Vetrina/Vetrina.purs
vapaj: And from the parent point of view, it should not even try to render vetrina if it knows something is wrong.

packages/vetrina/src/Vetrina/Vetrina.purs
vapaj: I mean, it would need to render the `UnexpectedServerError` view

packages/vetrina/src/Vetrina/Vetrina.purs
ae-mo: I used the `Maybe` to discriminate in the `didMount` hook. What different strategy would you suggest?

packages/vetrina/src/Vetrina/Vetrina.purs
```diff
@@ -253,6 +257,10 @@ render self = vetrinaContainer self $
             , productSelection: self.state.productSelection
             , onLogin: self.props.onLogin
             }
+        ServerError -> self.props.unexpectedError
```
vapaj, line 260: Let's only render it in case of `InitializationError` for now, as we are using e.g.`UnexpectedError` everywhere (where we would like to render the other error view instead), and address it when implementing issue #375

ae-mo: So for the record should I render the internal `vetrina` error view for cases `ServerError` and `UnexpectedError`?

vapaj: Yes, for now.

vapaj:
Would you also add a note of this prop to the [README](https://github.com/KSF-Media/affresco/blob/master/packages/vetrina/README.md) and bump the vetrina version, as this PR adds behavioural changes? 🙏

PR #376, opened 2020-05-15
integrated mover to fix timing problems

jockeF:
One less script. All ads functionality now starts from the same call and does not depend on DOM status.


scripts/app/gamAppAds.js
```diff
@@ -146,22 +209,28 @@ var ksfDfp = {};
...
+                        }, 200);
+                    }
                     var contentUnitDiv;
                     var headerToShow;
                     contentUnitDiv = event.slot.getSlotElementId();
 
                     // show ad headline fitting to the ad format. Out of page slots should have a headline that allows viewers to close the ad
                     // only show ad hedline for ads that are actually supposed to render
-                    contentUnitDiv = event.slot.getSlotElementId(); // this line is probably not needed?
+                    //                    contentUnitDiv = event.slot.getSlotElementId(); // this line is probably not needed?
```
vapaj, line 224: Rather than having commented code, can we just remove this line?

jockeF: Done

PR #378, opened 2020-05-20
WIP: add tracking for mitt-konto events

mfahler9:
added tracking to mitt-konto events (temporary address change, delivery reclamation, pause subscription)

### DIFF
--- a/apps/mitt-konto/src/Tracking.js
+++ b/apps/mitt-konto/src/Tracking.js
@@ -13,3 +13,21 @@ exports.pushPageLoad = function(tracker) {
     tracker.dataLayer.push({'event': 'page_data'});
   }
 }
+
+exports.reclamationEvent_ = function(tracker, subsno, date, action) {
+  return function() {
+      tracker.dataLayer.push({'event': 'reclamationEvent', 'subsno': subsno, 'date': date, 'action': action});
+  }
+}
+
+exports.tempAdressChange_ = function(tracker, subsno) {
+  return function() {
+      tracker.dataLayer.push({'event': 'tempAdressChangeEvent', 'subsno': subsno});
+  }
+}
+
+exports.pauseSubscription_ = function(tracker, subsno) {
+  return function() {
+      tracker.dataLayer.push({'event': 'pauseSubscriptionEvent', 'subsno': subsno});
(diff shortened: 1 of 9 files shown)


packages/components/src/DeliveryReclamation.purs
```diff
@@ -142,7 +146,12 @@ render self@{ state: { publicationDate, claim, maxPublicationDate }} =
           liftEffect $ self.props.onLoading
           User.createDeliveryReclamation self.props.userUuid self.props.subsno date'' claim'' >>=
             case _ of
-              Right recl -> liftEffect $ self.props.onSuccess recl
+              Right recl -> do
```
vapaj, line 149: We could remove the multiple `liftEffects` by having
```purescript
Right recl -> liftEffect do
...
```

apps/mitt-konto/src/Tracking.purs
```diff
@@ -3,10 +3,28 @@ module Tracking where
 import Effect (Effect)
 import Foreign (Foreign)
 import Prelude (Unit)
+import Data.Function.Uncurried (Fn2, Fn4, runFn2, runFn4)
 
 foreign import data Tracker :: Type
 foreign import newTracker :: Effect Tracker
 foreign import pushPageLoad :: Tracker -> Effect Unit
+foreign import reclamationEvent_ :: Fn4 Tracker Subsno DateString Claim (Effect Unit)
```
vapaj, line 11: I think it would be more appropriate to use [EffectFn](https://pursuit.purescript.org/packages/purescript-effect/2.0.1/docs/Effect.Uncurried) here, as all these are effectful functions. So for instance this would become `EffectFn4 Tracker Subsno DateString Claim Unit`

apps/mitt-konto/src/Tracking.js
```diff
@@ -13,3 +13,21 @@ exports.pushPageLoad = function(tracker) {
     tracker.dataLayer.push({'event': 'page_data'});
   }
 }
+
+exports.reclamationEvent_ = function(tracker, subsno, date, action) {
+  return function() {
```
vapaj, line 18: I don't think we need the wrapping function 🤔

PR #384, opened 2020-06-04
Fix netlify serverless functions for podcasts app

joakimekroos:
Moved the `netlify-lambda build` task from netlify.toml to package.json


joakimekroos:
That actually did the trick. [Deploy preview](https://5ed8b646a8435f0007b829b3--ksf-podcasts.netlify.app/)  on Netlify works

PR #388, opened 2020-06-15
Transaction tracking

mfahler9:


packages/vetrina/src/Tracking.purs
```diff
@@ -0,0 +1,14 @@
...
+import Data.Maybe (Maybe(..))
+import Effect (Effect)
+import KSF.User (OrderNumber)
+import Prelude (Unit, pure, unit)
+
+foreign import data Tracker :: Type
+foreign import transaction_ :: Fn3 OrderNumber String String (Effect Unit)
+
+transaction :: OrderNumber -> Maybe String -> Maybe String -> Effect Unit
+transaction ord (Just t) (Just p)  = runFn3 transaction_ ord t p
```
vapaj, line 13: Instead of calling these `t` and `p`, let's make up something more meaningful.

vapaj: `ord` could also be better (this would imply it's maybe an order of some sorts?), however one can see from the type signature it's an `OrderNumber` so it's fine-ish 🙂 The other ones are just `Strings` so I have no idea what they are.

packages/vetrina/src/Vetrina/Vetrina.purs
```diff
@@ -202,6 +204,9 @@ pollOrder setState state@{ logger } (Right order) = do
           nextPurchaseStep = case userAccountStatus of
             NewAccount      -> PurchaseSetPassword
             _               -> PurchaseCompleted userAccountStatus
+      product_id <- liftEffect $ LocalStorage.getItem "product_id" --analytics
+      product_price <- liftEffect $ LocalStorage.getItem "product_price" --analytics
```
vapaj, line 208: Do we want to send a warning to sentry if `product_id` and `product_price` are not found in the local storage? Do we care?

BTW let's not mix `snake_case` with `camelCase`, so rename these to `productId` and `productPrice` (however it's fine to have `"product_price"` in the local storage if you want)

Issue #396, opened 2020-07-03
Improve pause subscription functionality


ae-mo:
We should:

1. Add the following text below the subscription pauses list in a subscription box:

   > Om du vill ändra på din tillfälliga adressändring eller ditt uppehåll, vänligen radera den gamla först och lägg sedan in den nya.

1. Make the first subscription pause starting date available in the calendar widget the day after tomorrow. We do this because we suspect that our backend might be preventing the creation of pauses after noon of the day before the desided starting date.

1. Reconsider the policy for too close pauses. Who decided it should be the way it is? Why?

ae-mo:
Point #2 is already good.

ae-mo:
It remains only point 3, but maybe this shouldn't be discussed here.

PR #397, opened 2020-07-13
Fix/pause subscription

ae-mo:
Related to #396.

### DIFF
--- a/packages/components/src/Subscription/Component.purs
+++ b/packages/components/src/Subscription/Component.purs
@@ -113,7 +113,7 @@ render self@{ props: props@{ subscription: sub@{ package } } } =
              , { term: "Status:"
                , description:
                    [ DOM.text $ translateStatus props.subscription.state ]
-                   <> (map DOM.text $ foldMap (showPausedDates <<< filterExpiredPausePeriods) $ self.state.pausedSubscriptions)
+                   <> (map DOM.text $ foldMap (addPauseDescription <<< showPausedDates <<< filterExpiredPausePeriods) $ self.state.pausedSubscriptions)
                }
              ]
              <> deliveryAddress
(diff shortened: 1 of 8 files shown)


packages/vetrina/package.json
```diff
@@ -4,21 +4,22 @@
   "description": "Vetrina, frontend for purchasing KSF Media products",
   "license": "MIT",
   "scripts": {
-    "start": "parcel index.html -p 8000",
-    "build": "spago build && parcel build index.html --no-cache"
```
f-f, line 8: No more `start` and `build` scripts? :thinking:

packages/components/src/Subscription/Component.purs
```diff
@@ -382,6 +386,14 @@ render self@{ props: props@{ subscription: sub@{ package } } } =
     billingPeriodEndDate =
           map trim $ formatDate =<< toMaybe props.subscription.dates.end
 
+    -- NOTE: We have a rule in our company policy that states that subscription pauses should be 7 days apart.
+    -- Thus, if a customer wants to extend a pause, they can't do it by adding a new pause immediately after it.
+    -- This is why we tell the customer to delete the pause and create a new one.
+    pauseDescription = DOM.div
+                         { className: "mitt-konto--note"
+                         , children: [ DOM.text "Om du vill ändra på din tillfälliga adressändring eller ditt uppehåll, vänligen radera den gamla först och lägg sedan in den nya." ]
+                         }
```
f-f, line 395: This is just temporary before we implement backend magic to that users don't have to care about this, right?
Or is this meant to be permanent?

ae-mo: We have no idea at the moment, first we have to understand our company policy better 😄

packages/vetrina/package.json
ae-mo: I have no idea why this changes, need to revert them

ae-mo: Ah no wait, I know why: I ran `yarn install` locally and it must have figured out that these scripts are unnecessary, because they were moved to the `package.json` of the `vetrina-staging` app. We now use that to produce local test builds.

Issue #290, opened 2020-03-17
Think about monorepo deploy


vapaj:
Currently, apps in our monorepo are deployed only when there are changes under the app directory. We have the `.build-size-sha256` files to trigger deploys, but we need to change these by hand (by hand I mean triggering [this perl script](https://github.com/KSF-Media/affresco/blob/master/hooks/pre-commit)), and it's hard to remember.

I wish there were a better way of doing this.

ae-mo:
Can't we make the script create an additional commit containing the updated hashes?

ae-mo:
I think this is outdated @vapaj ?

vapaj:
Indeed 🤠

PR #402, opened 2020-07-23
Improve monorepo deploy

vapaj:
Improves #290

Changes:
- commits build trigger files automatically
- builds spago projects before calculating trigger hashes (because it's important to have up to date hashes, otherwise they would always change). The downside to this is that pre-commits would take a lot of time, which is why...
- uses `pre-push` hook instead of `pre-commit` hook, because we don't need to create trigger files that often really

Let's see if this makes developing affresco any less annoying (it should) 💩


vapaj:
This is not that urgent PR, so @ae-mo, let's go through this when the holidays are over 🙂

f-f:
Maybe I'm missing something, but what's the reason why we need to have the hash files?

vapaj:
> Maybe I'm missing something, but what's the reason why we need to have the hash files?

@f-f It's because of the Netlify monorepo deploys: Netlify only deploys apps when there changes under a configured app directory, which is neat. However, this gets tricky with our purescript apps, as the changes are mostly made outside of the app directory. To tackle this, we have these deploy trigger files inside the app directories, which simply are the sha256 hashes of the content of the complied purescript of the `/output` dir.

What initially sounded a like simple thing, is becoming a perl monster.

f-f:
Yeah I had that feeling.

I think there are two most sensible paths here:
1. just redeploy everything when anything changes. Questions about that:
  - is it even possible to configure Netlify to do it?
  - if yes, is that too slow or using too many CI minutes?
2. use a smarter (and most importantly already implemented) build system to do this - would Make work here if we use it to run all the deployments? I think it should be able to pickup changes if we define targets properly?

vapaj:
> is it even possible to configure Netlify to do it?

Yes, it was initially like this

> if yes, is that too slow or using too many CI minutes?

It's super slow and often hangs/cancels deployments because of too many concurrent jobs (also doesn't make sense to upgrade our plan just for this)

> would Make work here

Possibly. Do you have any concrete ideas on this?

vapaj:
Does @mrskug have any ideas?

Basically every solution not adding stuff to git would be preferable.

ae-mo:
Sorry guys I completely missed that I was appointed as reviewer 😅

f-f:
@ae-mo I think this is going to be replaced by #404?

PR #409, opened 2020-08-13
Show sub actions on success

### LINKED ISSUE #408: Do not hide subscription actions after an action is perfomed
When the user performs and action on a subscription, we display either a confirmation or an error text.

In both cases, we hide the action buttons, so that if the user wants to perform another action on the subscription, they have to reload the page or log out and log in again.

We should change this so that the action buttons are still visible after an action is performed.

mrskug:
Fixes #408

### DIFF
--- a/packages/components/src/Subscription/Component.purs
+++ b/packages/components/src/Subscription/Component.purs
@@ -167,18 +167,29 @@ render self@{ props: props@{ subscription: sub@{ package } } } =
         where
           asyncWrapper = AsyncWrapper.asyncWrapper
             { wrapperState: self.state.wrapperProgress
-            , readyView: actionsContainer $ [ pauseIcon
-                                            , if maybe true Array.null self.state.pausedSubscriptions
-                                              then mempty
-                                              else removeSubscriptionPauses
-                                            , temporaryAddressChangeIcon
-                                            , deliveryReclamationIcon
-                                            ]
+            , readyView: actionsContainer $ defaultActions
             , editingView: identity
-            , successView: \msg -> successContainer [ DOM.div { className: "subscription--update-success check-icon" }, foldMap successMessage msg  ]
+            , successView: \msg -> actionsContainer $ append defaultActions $ [successWrapper msg]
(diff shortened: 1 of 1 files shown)


packages/components/src/Subscription/Component.purs
```diff
@@ -167,19 +167,30 @@ render self@{ props: props@{ subscription: sub@{ package } } } =
...
-                                              then mempty
-                                              else removeSubscriptionPauses
-                                            , temporaryAddressChangeIcon
-                                            , deliveryReclamationIcon
-                                            ]
+            , readyView: actionsContainer $ defaultActions
             , editingView: identity
-            , successView: \msg -> successContainer [ DOM.div { className: "subscription--update-success check-icon" }, foldMap successMessage msg  ]
+            , successView: \msg -> actionsContainer $ append defaultActions $ [successWrapper msg]
             , errorView: \err -> errorContainer [ errorMessage err, tryAgain ]
```
ae-mo, line 173: As far as I understood, we need to show the actions in case of error as well.

mrskug: Ah true, see latest commit ;)

packages/components/src/Subscription/Component.purs
```diff
@@ -167,19 +167,33 @@ render self@{ props: props@{ subscription: sub@{ package } } } =
...
-                                              then mempty
-                                              else removeSubscriptionPauses
-                                            , temporaryAddressChangeIcon
-                                            , deliveryReclamationIcon
-                                            ]
+            , readyView: actionsContainer $ defaultActions
             , editingView: identity
-            , successView: \msg -> successContainer [ DOM.div { className: "subscription--update-success check-icon" }, foldMap successMessage msg  ]
-            , errorView: \err -> errorContainer [ errorMessage err, tryAgain ]
+            , successView: \msg -> actionsContainer $ append defaultActions $ [successWrapper msg]
```
ae-mo, line 172: Just occurred to me, you can use `<>` instead of `append`, it is more common in our codebase and in general 😃

PR #410, opened 2020-08-17
Add payment method

mfahler9:
Show payment method for each subscription

### DIFF
--- a/packages/components/src/Api/Subscription.purs
+++ b/packages/components/src/Api/Subscription.purs
@@ -37,10 +37,10 @@ type Subscription =
   , paused                :: Nullable (Array PausedSubscription)
   , deliveryAddress       :: Nullable DeliveryAddress
   , pendingAddressChanges :: Nullable (Array PendingAddressChange)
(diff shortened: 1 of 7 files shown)


packages/components/src/Api/Subscription.purs
```diff
@@ -37,8 +37,10 @@ type Subscription =
   , paused                :: Nullable (Array PausedSubscription)
   , deliveryAddress       :: Nullable DeliveryAddress
   , pendingAddressChanges :: Nullable (Array PendingAddressChange)
+  , paymentMethod         :: SubscriptionPaymentMethod
   }
 
+newtype SubscriptionPaymentMethod = SubscriptionPaymentMethod String
```
vapaj, line 43: Instead of a `newtype`, let's make this into a sum type, because when we'll start adding features concerning the payments, it'll come handy.

```purescript
data SubscriptionPaymentMethod
  = PaperInvoice
  | CreditCard
  | NetBank
  ...
```

vapaj (approved):

PR #413, opened 2020-09-01
Allow duration of temporary address change to be seven days

### LINKED ISSUE #411: Allow duration of temporary address change to be seven days
In Mitt Konto, the minimum duration for a temporary address change for a subscription is supposed to be seven days, but due to a small mistake in our code, it is actually eight days. Here is the incriminated code:

https://github.com/KSF-Media/affresco/blob/7fd4f520db90368668e3ae1fc343052da0be8dec/packages/components/src/TemporaryAddressChange/Component.purs#L98-L100

We should fix this so that the minimum duration can actually be seven days, because the way we do it we add seven days to the starting date, which makes a total of eight days.

mrskug:
Fixes #411

### DIFF
--- a/packages/components/src/TemporaryAddressChange/Component.purs
+++ b/packages/components/src/TemporaryAddressChange/Component.purs
@@ -96,6 +96,7 @@ component = React.createComponent "TemporaryAddressChange"
 calcMinEndDate :: Maybe DateTime -> Maybe DateTime
 calcMinEndDate Nothing = Nothing
 calcMinEndDate (Just startDate) = do
+  -- 6 days added to the starting date = 7 (one week)
   let week = Time.Duration.Days 6.0
   adjust week startDate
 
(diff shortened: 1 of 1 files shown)


packages/components/src/TemporaryAddressChange/Component.purs
```diff
@@ -96,7 +96,7 @@ component = React.createComponent "TemporaryAddressChange"
 calcMinEndDate :: Maybe DateTime -> Maybe DateTime
 calcMinEndDate Nothing = Nothing
 calcMinEndDate (Just startDate) = do
-  let week = Time.Duration.Days 7.0
+  let week = Time.Duration.Days 6.0
```
ae-mo, line 99: Technically this is not a week anymore 😆 maybe you can call it something like `interval` or `days` (I don't  really know) and add a comment on why it's `6.0`.

mrskug: I think `week` is a good name here as that is the result of the action we are doing here and conveys the intent better than just `interval` or `days`

ae-mo: I'd rather add a comment to explain that we want the minimum duration to be one week

pikkuruusu: So, could we find a solution here? Customer service is anxious.

mrskug: Feel free to merge at any time

ae-mo (approved):

PR #416, opened 2020-09-09
In text ad refactoring

jockeF:
Some major changes to how in text ads are handled. Some code in App.js was removed and replaced with a function that checks paragraphs to see if they are appropriate for ad placement. There are two ad slots placed, DIGIHELMOB and MOBMITT. The MOBMITT was mostly failing to get placed with the old code. The code placing DIGIHELMOB was previously in the external ads script and suffered from some timing issues. The placement code is now removed from the external script and that script is now run after the slots have been placed, from App.js.
Note that both the App.js and the external ad script must be updated for this to work, so caching issues might cause problems.


joakimekroos (approved):
Not tested locally, but spotted no errors in code.

PR #417, opened 2020-09-10
Indefinite temp addr change

mrskug:


packages/components/src/InputSelect/Component.purs
```diff
@@ -0,0 +1,99 @@
+module KSF.InputSelect.Component where
```
ae-mo, line 1: You can avoid the old convention of creating a folder for each component and put the code directly in a file called `InputSelect.purs`

packages/components/src/InputSelect/Component.purs
```diff
@@ -0,0 +1,99 @@
...
+import React.Basic.DOM.Events (targetChecked)
+import React.Basic.Events (handler)
+
+type Self = React.Self Props State
+
+type Props =
+  { type_           :: InputType
+  , name            :: String
+  , value           :: Maybe String
+  , checked         :: Maybe Boolean
```
ae-mo, line 24: Why does it need to be wrapped in a `Maybe`?

packages/components/src/InputSelect/Component.purs
```diff
@@ -0,0 +1,99 @@
...
+
+component :: React.Component Props
+component = React.createComponent "InputSelect"
+
+inputSelect :: Props -> JSX
+inputSelect = React.make component
+  { render, didMount, initialState }
+  where
+    didMount { props, setState } = when (isJust props.checked) $
+      setState \s -> s { checked = fromMaybe false props.checked }
```
ae-mo, line 47: I think this bit could be improved. You are calling `setState` when `props.checked` contains something, but then you use `fromMaybe` on it. I think this is an undesired consequence of having `checked` wrapped in a `Maybe`. Can't it be a simple `Boolean` who is assigned `False` when instantiating the component?

packages/components/src/Persona.purs
```diff
@@ -132,11 +132,13 @@ temporaryAddressChange
   -> Aff Subscription
 temporaryAddressChange uuid subsno startDate endDate streetAddress zipCode countryCode temporaryName token = do
   let startDateISO = formatDate startDate
-      endDateISO   = formatDate endDate
+      endDateISO   = case endDate of
```
ae-mo, line 135: `Maybe a` is an instance of the `Functor` typeclass, that is, it can be mapped over, meaning that if you want to get a `Maybe b` value from a `Maybe a` value by applying a function, you can use `fmap`, aka `<$>`.

I suggest you change this to
```
formatDate <$> endDate
```

packages/components/src/InputSelect/Component.purs
ae-mo: And maybe we could think of a better name, cause `InputSelect` remembers me of the HTML `select` tag and this might generate confusion.

packages/components/src/TemporaryAddressChange/Component.purs
```diff
@@ -29,14 +30,15 @@ import React.Basic.DOM.Events (preventDefault)
...
-  , temporaryName :: Maybe String
+  { startDate      :: Maybe DateTime
+  , minStartDate   :: Maybe DateTime
+  , endDate        :: Maybe DateTime
+  , minEndDate     :: Maybe DateTime
+  , streetAddress  :: Maybe String
+  , zipCode        :: Maybe String
+  , countryCode    :: Maybe String
+  , temporaryName  :: Maybe String
+  , perpetualCheck :: Maybe Boolean
```
ae-mo, line 41: I think we can get rid of the `Maybe` here as well. 

As for the name of this property, I would change it to something like `isIndefinite`.

packages/components/src/InputSelect/Component.purs
```diff
@@ -0,0 +1,99 @@
...
+  , checked         :: Maybe Boolean
+  , onChange        :: Maybe Boolean -> Effect Unit
+  , label           :: Maybe String
+  , required        :: Maybe Boolean
+  }
+
+type State =
+  { checked :: Boolean }
+
+data InputType = Radio | Checkbox
```
vapaj, line 33: I guess the idea is to use this component also for the `Radio` inputs we now use via `InputField` component, right? Why did you decide to separate this (and `Radio`) from the `InputField` component?

packages/components/src/InputSelect/Component.purs
vapaj: If we want to separate the "selectable" input types into their own module, it would make sense to place it under `KSF.InputField`. So it would be named something like `KSF.InputField.Selectable`. This way we can still treat text inputs and boolean inputs as `InputFields` on the outside (if we want), and the type `data InputType = Text | Password | Email | Radio | Checkbox` would stay coherent.

packages/components/src/TemporaryAddressChange/Component.purs
mrskug: I like the name but this needs to be a Maybe as that is what `React.Basic.DOM.Events.targetChecked` returns

packages/components/src/InputSelect/Component.purs
mrskug: If I change the InputField component to support checkboxes that requires adding a prop for checked that is totally useless but still required to be defined for all other input types. As radio buttons and checkboxes have similar uses It makes sense to separate them from the rest of the input types. The Radio part of this component is not implemented fully yet so I might just remove it from here and it could be added if we start using radio buttons more in the future.

packages/components/src/TemporaryAddressChange/Component.purs
vapaj: We can assume that `Nothing` means  `false` here (what we actually do in other parts of the component already).

Seems you are already looking into the return value of `targedChecked`:
```purescript
onChange: handler targetChecked \maybeNewVal -> do
                self.setState _ { checked = fromMaybe false maybeNewVal }
                props.onChange maybeNewVal
```
What's missing is to change `onChange` to take `Boolean` instead of `Maybe Boolean`.

vapaj: Or handle the `Maybe` in this part
```purescript
onChange: \checked -> self.setState _ { perpetualCheck = checked }
```

packages/components/src/Persona.purs
mrskug: 4d7c964

packages/components/src/TemporaryAddressChange/Component.purs
mrskug: Sense this makes. cc742f8

packages/components/src/InputSelect/Component.purs
mrskug: `KSF.InputField.Selectable` makes a lot of sense. Should this then be grouped under the InputField folder?

vapaj: Preferably yes.

packages/components/src/InputField/Checkbox.purs
```diff
@@ -0,0 +1,81 @@
+module KSF.InputField.Checkbox where
```
mrskug, line 1: Just going to call this module checkbox. Trying to make radio buttons and checkboxes behave nicely and logically within the same module is not feasible imo. They are just too different in their behaviour.

packages/components/src/InputSelect/Component.purs
mrskug: continued here https://github.com/KSF-Media/affresco/pull/417#discussion_r487782364

apps/vetrina-staging/.build-size-sha256
```diff
@@ -0,0 +1,3 @@
+# SHA256 of the content of the build directory.
```
vapaj, line 1: This can be removed as it's not in the repo anymore

packages/components/src/TemporaryAddressChange/Component.purs
```diff
@@ -161,14 +164,25 @@ render self@{ state: { startDate, endDate, streetAddress, zipCode, countryCode,
         , label: "Börjar från"
         }
 
+    isIndefiniteCheckbox =
+      InputCheckbox.inputCheckbox
+        { type_: InputCheckbox.Checkbox
+        , name: "indefinite"
+        , value: Nothing
+        , checked: false
```
vapaj, line 172: Should this be `state.isIndefinite`? Always setting it `false` implies that whenever the state changes (and this thing is re-rendered), the checkbox gets unchecked.

vapaj: Oh, you don't actually use it for anything in `InputCheckbox`. Or am I missing something?

mrskug: :thinking:

vapaj (approved):

PR #419, opened 2020-09-14
Vetrina/multiple products

vapaj:
Adds multiple product support to Vetrina.

Also removes the `vetrina-staging` js app and adds `vetrina-test` purs app, just to make life a bit easier when developing this thing.


apps/vetrina-test/packages.dhall
```diff
@@ -0,0 +1,111 @@
+{-
```
ae-mo, line 1: I guess we can get rid of this mini guide

apps/vetrina-test/spago.dhall
```diff
@@ -0,0 +1,29 @@
+{-
```
ae-mo, line 1: Same as above

apps/vetrina-test/packages.dhall
vapaj: Don't know if it matters that much. It also includes nice documentation if someone (new to spago and purescript) wonders what that is.

apps/vetrina-test/spago.dhall
vapaj: The name of the project should be at least fixed 😅

apps/vetrina-test/packages.dhall
vapaj: Compromise: We can just remove this file :D

packages/vetrina/src/Vetrina/Products.purs
```diff
@@ -0,0 +1,63 @@
...
+render self =
+  DOM.div
+    { className: "products--container"
+    , children:  map renderProduct self.props.products
+    }
+  where
+    renderProduct p =
+      DOM.label
+        { className: "vetrina--product"
+        , children: [ DOM.input $
```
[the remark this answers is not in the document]
vapaj: The `InputField` component exists mainly because of the styling. The products we show in Vetrina are very unique in design.

Should have attached this image in the description... Anyway, this is roughly what we want (only some texts and the background color of the button have changed.)
image (3)

ae-mo (approved):

Issue #422, opened 2020-09-15
Replace loading spinner with appropriate message when redirecting to Nets


vapaj:
Currently we are showing a loading spinner when we're redirecting the user to the Nets terminal. Let's instead have a descriptive message there to inform the customer what's going on.

@pikkuruusu knows what to show here 😊

PR #425, opened 2020-09-15
Display a message when redirecting to Nets in Vetrina (#422)

kaol:
* Add a function Spinner.loadSpinnerWithMessage which displays a message along with the spinner

### DIFF
--- a/packages/components/src/Spinner.purs
+++ b/packages/components/src/Spinner.purs
@@ -55,5 +55,13 @@ loadingSpinner =
         ]
     }
 
+loadingSpinnerWithMessage :: String -> JSX
+loadingSpinnerWithMessage msg =
+  DOM.div
+    { className: "spinner--loading-spinner"
+    , children: [ DOM.text msg ]
+    }
+  <> loadingSpinner
+
 setSpinner :: forall r. Maybe Loading -> { isLoading :: Maybe Loading | r } -> { isLoading :: Maybe Loading | r }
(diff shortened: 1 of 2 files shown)


packages/components/src/Spinner.purs
```diff
@@ -55,5 +55,18 @@ loadingSpinner =
         ]
     }
 
+loadingSpinnerWithMessage :: String -> JSX
```
vapaj, line 58: The alignment of the spinner and the message breaks a bit here: the spinner seems to be aligned left, although it should be centered. As the `.spinner--loading-spinner` is a [grid](https://css-tricks.com/snippets/css/complete-guide-grid/), we can adjust the alignment by setting [justify-items](https://developer.mozilla.org/en-US/docs/Web/CSS/justify-items) to `center`.

I don't think the `align-content` and `justify-content` attributes, which we have defined in that class, are actually useful anymore.

kaol: I didn't touch spinner's styles but instead added `.spinner--loading-spinner` to the div containing the message and made it a sibling of the div containing the spinner. With this, both are justified to the center and the message is on top of the spinner. I'll change it if you have a better idea about how it should look like, this was what I was aiming for initially but I used a too short string for testing.

vapaj: Yep, this works! 👍

vapaj (approved):

Issue #412, opened 2020-08-31
Update credit card information


ae-mo:
In **Mitt Konto**, the user should be able to update credit card information, in order to avoid the abrupt ending of their subscription(s).

PR #433, opened 2020-09-28
Fix/update bottega

ae-mo:
Related to #412.

### DIFF
--- a/packages/components/src/Bottega/Models/CreditCard.purs
+++ b/packages/components/src/Bottega/Models/CreditCard.purs
@@ -21,14 +21,14 @@ type CreditCard =
 
 newtype CreditCardRegisterNumber = CreditCardRegisterNumber String 
 
-data CreditCardRegister = CreditCardRegister
+type CreditCardRegister =
   { number      :: CreditCardRegisterNumber
   , user        :: UUID
   , terminalUrl :: Maybe PaymentTerminalUrl
   , status      :: CreditCardRegisterStatus
   }
 
-data CreditCardRegisterStatus = CreditCardRegisterStatus
+type CreditCardRegisterStatus =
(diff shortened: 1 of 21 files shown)


packages/components/src/Bottega/Models/CreditCard.purs
```diff
@@ -0,0 +1,52 @@
...
+import Data.Maybe (Maybe, maybe)
+import KSF.Api (UUID)
+
+newtype CreditCardId = CreditCardId Int
+
+type CreditCard =
+  { id              :: CreditCardId
+  , user            :: UUID
+  , paymentMethodId :: Maybe PaymentMethodId
+  , panHash         :: Maybe String
```
vapaj, line 17: Actually, can this ever be missing in the UI?

vapaj: Hmm, right, as `POST /paymentMethod/creditCard` creates an empty credit card

vapaj: However from the point of view of just getting the users credit cards, I think the data is invalid if these fields are missing.

ae-mo: So we should throw an error?So we should throw an error?

vapaj: So for example when showing the cards to the user, we would need to validate them client side.

We could have a concept of `PendingCreditCard` or `EmptyCreditCard` which would not include these fields. This would also solve the problem of having an optional body in the `POST /paymentMethod/creditCard` endpoint, as then it would always be required. But let's think about this in another time.

vapaj (approved):

PR #445, opened 2020-10-21
Add a dropdown menu for selecting a previously used temporary address

### LINKED ISSUE #406: Retrieve existing addresses for temporary address changes
A user should be able to select an address that they previously used for a temporary address change.

kaol:
Closes #406

### DIFF
--- a/packages/components/src/Persona.purs
+++ b/packages/components/src/Persona.purs
@@ -347,6 +347,7 @@ type User =
   , subs :: Array Subscription
   , consent :: Array GdprConsent
   , pendingAddressChanges :: Nullable (Array PendingAddressChange)
+  , pastTemporaryAddresses :: Array TemporaryAddressChange
   , hasCompletedRegistration :: Boolean
   }
 
@@ -373,6 +374,14 @@ type Address =
   , houseNo       :: Nullable String
   , staircase     :: Nullable String
   , apartment     :: Nullable String
+  }
+
+type TemporaryAddressChange =
+  { street        :: String
+  , zipcode       :: String
+  , cityName      :: Nullable String
+  , countryCode   :: String
+  , temporaryName :: Nullable String
   }
 
 type GdprConsent =
--- a/packages/components/src/TemporaryAddressChange/DropDown.purs
+++ b/packages/components/src/TemporaryAddressChange/DropDown.purs
@@ -1,0 +1,47 @@
+module KSF.TemporaryAddressChange.DropDown where
+
+import Prelude (Unit, map, show, ($), (<$>), (>>>), (<<<), (<>))
+
+import Control.Bind
+import Data.Enum (enumFromTo)
+import Data.Function (flip)
+import Data.Int as Int
+import Data.Foldable (class Foldable)
+import Data.List (zip, length, fromFoldable, filter)
+import Data.Map as Map
+import Data.Maybe (Maybe, isJust, fromMaybe)
+import Data.Tuple (Tuple (..))
+import Effect (Effect)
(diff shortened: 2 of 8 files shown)


packages/components/src/TemporaryAddressChange/DropDown.purs
```diff
@@ -0,0 +1,47 @@
...
+    { className: "input-field--container"
+    , style: DOM.css { "width": "100%" }
+    , children:
+        [ inputLabel { label: "Välj adress", nameFor: "pastTemp" }
+        , DOM.select
+            { id: "pastTemp"
+            , children:
+                [ DOM.option { value: "-", children: [ DOM.text "Ny adress" ] } ] <>
+                (map createOption $ Map.toUnfoldable $ (fromMaybe "" <<< _.streetAddress) <$> addrIndexed)
+            , onChange: handler (targetValue >>> unsafeEventFn setAddress) onChange
```
vapaj, line 38: This could be `handler targetValue (setAddress >>> onChange)`

apps/mitt-konto/package.json
```diff
@@ -9,7 +9,9 @@
     "less": "^3.8.1",
     "react": ">=15.5",
     "react-date-picker": "^7.5.0",
-    "react-dom": ">=15.5"
+    "react-dom": ">=15.5",
+    "spago": "^0.15.2",
```
vapaj, line 13: We probably don't need `spago` or `yarn` here right? 🤔

kaol: `yarn upgrade` or something must have automatically added them, I know I didn't do it myself.

packages/components/src/TemporaryAddressChange/DropDown.purs
```diff
@@ -0,0 +1,47 @@
...
+  DOM.div
+    { className: "input-field--container"
+    , style: DOM.css { "width": "100%" }
+    , children:
+        [ inputLabel { label: "Välj adress", nameFor: "pastTemp" }
+        , DOM.select
+            { id: "pastTemp"
+            , children:
+                [ DOM.option { value: "-", children: [ DOM.text "Ny adress" ] } ] <>
+                (map createOption $ Map.toUnfoldable $ (fromMaybe "" <<< _.streetAddress) <$> addrIndexed)
```
vapaj, line 37: Probably not a good idea show empty options in the dropdown if streetAddress is missing.

packages/components/src/Persona.purs
```diff
@@ -375,6 +376,14 @@ type Address =
   , apartment     :: Nullable String
   }
 
+type TemporaryAddressChange =
```
vapaj, line 379: The Persona `PastTemporaryAddress` has the type
```
countryCode* | string
zipcode* | string
cityName | string
street* | string
temporaryName | string
```

Should we mimic that here?

vapaj: Meaning, by Personas standards, `countryCode`, `zipCode` and `street` should always exist.

packages/components/src/TemporaryAddressChange/DropDown.purs
kaol: There's a `filter (isJust <<< _.streetAddress) $ fromFoldable pastAddresses` at the start of the function so it's not supposed to ever happen.

packages/components/src/Persona.purs
kaol: I made the change, though I ended up using `Just` for them when I translated them to `AddressChange` type. I think we're better of sticking to `Maybe` for that type's fields since that reflects how input fields store their values in them.

vapaj (approved):

PR #446, opened 2020-10-22
Fix form fields not aligning nicely

vapaj:
This was because of basscss for some weird reason


kaol (approved):
I would have squashed those three commits.

vapaj:
We do squash at merge though

PR #447, opened 2020-10-22
Payment history view

### LINKED ISSUE #407: Show the history of transactions
Show the history of transactions made by the customer in `mitt-konto`. 

Probably the way to go would be to open a modal containing a paged list. 

IMPORTANT: we need to show tax information for each transaction (actual numbers!).

kaol:
Brings router as a dependency.

Closes #407


apps/mitt-konto/src/MittKonto/Main.purs
```diff
@@ -306,6 +336,11 @@ userView { setState, state: { logger } } user = React.fragment
           , description: "Byt lösenord"
           , className: passwordChangeClass
           }
+      , formatIconLink
+          { href: "/fakturor"
```
vapaj, line 340: I still think it would be better to open this in the same the tab, as it's going to load a second instance of mitt konto. Now that we're using react router, even the browser history works as expected.

packages/components/src/PaymentAccordion/Component.purs
```diff
@@ -0,0 +1,170 @@
+module KSF.PaymentAccordion.Component where
```
vapaj, line 1: Good stuff!

packages/components/src/Persona.purs
```diff
@@ -434,3 +437,116 @@ instance writeForeignDeliveryReclamationStatus :: WriteForeign DeliveryReclamati
...
+  , interest :: Number
+  , vat :: Number
+  , amount :: Number
+  , openAmount :: Number
+  , type :: PaymentType
+  , state :: PaymentState
+  , discount :: Number
+  }
+
+type SubscriptionPayments' =
```
vapaj, line 501: Very nice to separate the js/json type from the type we're using in code. We should do this for other types as well where it makes sense. However, could we call this `ApiSubscriptionPayments` or `JSSubscriptionPayments` or `SubscriptionPaymentsJS` or something similar? We do have e.g. `JSProps` in some of our react components, so `JS` prefix or postfix would be in line with those 🤔

packages/components/src/PaymentAccordion/Component.purs
```diff
@@ -0,0 +1,170 @@
...
+  }
+
+didMount :: Self -> Effect Unit
+didMount self = do
+  pure unit
+
+render :: Self -> JSX
+render self =
+  DOM.div
+    { className: "mitt-konto--payments"
```
vapaj, line 52: It might be a good idea to separate the css of this particular component into its own file, as it's now dependent on mitt konto. The convention is just to create a `<ComponentName>.less` under the `/less` directory.

packages/components/src/PaymentAccordion/Component.purs
vapaj, line 1: Oh and we can just move this file right under `/components/src` and call it `KSF.PaymentAccordion`. The `ComponentName/Component.purs` is a convention we're moving out from.

apps/mitt-konto/src/MittKonto/Main.purs
```diff
@@ -362,6 +401,18 @@ loginView self@{ state: state@{ logger }, setState } = React.fragment
               Right user -> do
                 setState $ setLoggedInUser $ Just user
                 logger.setUser $ Just user
+                path <- pathname =<< location =<< window
+                case path of
+                  "/fakturor" -> Aff.launchAff_ $ withSpinner (setState <<< setLoading) do
```
vapaj, line 406: I don't really like pattern matching on the location when react router kind of does that for us already. Of course the problem here is where to fetch the payments.

How do you feel about passing `Aff (Array SubscriptionPayments)` to `PaymentAccordion` props and fetching them in the `didMount` function there? 🤔

packages/components/src/Persona.purs
kaol: I changed it to have `Api` prefix. It's not used outside of this file and I wouldn't expect it to be either.

packages/components/src/PaymentAccordion/Component.purs
kaol: Done.

packages/components/src/PaymentAccordion/Component.purs
kaol: Done.

apps/mitt-konto/src/MittKonto/Main.purs
kaol: I moved it to `paymentsView` and used `didMount` and it's much nicer this way.

less/PaymentAccordion.less
```diff
@@ -0,0 +1,42 @@
+.mitt-konto--payment-header {
```
vapaj, line 1: Right. The convention is also to name the classes `.<component-name>--class-name` so for instance `.payment-accordion--payment-header` here 🙂 It makes inspecting the DOM so much nicer.

kaol: Ok, I updated the class names. Though I loaned `mitt-konto--component-block-content` for my styling and I didn't copy that under a different name.

vapaj (approved):

PR #449, opened 2020-10-27
Feature/credit card update

### LINKED ISSUE #412: Update credit card information

ae-mo:
Related to #412.


ae-mo:
It's a draft because:

- it's full of unused imports

- style needs improvement, css classes need correct names

- I need to make a better job hiding the loading of a user's credit cards

- I need to hide the ugly nets iframe during processing behind a loader

- etc.

packages/components/src/Bottega.purs
```diff
@@ -50,10 +51,17 @@ readCreditCard :: { id :: CreditCardId, user :: UUID, paymentMethodId :: Payment
 readCreditCard creditCardObj = pure $ 
   { id: creditCardObj.id, user: creditCardObj.user, paymentMethodId: creditCardObj.paymentMethodId, maskedPan: creditCardObj.maskedPan, expiryDate: creditCardObj.expiryDate }
 
-readCreditCardRegister :: { number :: CreditCardRegisterNumber, user :: UUID, terminalUrl :: Nullable PaymentTerminalUrl, status :: { state :: String, time :: String, failReason :: Nullable String }  } -> Aff CreditCardRegister
+readCreditCardRegister :: { number :: CreditCardRegisterNumber, user :: UUID, creditCardId :: CreditCardId, paymentTerminalUrl :: Nullable PaymentTerminalUrl, status :: { state :: String, time :: String, failReason :: Nullable String }  } -> Aff CreditCardRegister
```
vapaj, line 54: A minor note: Instead of inlining this type, we could separate it to e.g. `ApiCreditCardRegister` as we are doing with the payment types: https://github.com/KSF-Media/affresco/blob/master/packages/components/src/Persona.purs#L510-L528. 
I think this is a good convetion for a better future.

less/CreditCard.less
```diff
@@ -0,0 +1,25 @@
+.credit-card-update--register-wrapper {
+    display: flex;
+    align-items: center;
+    justify-content: center;
+}
+
+.credit-card-update--register-terminal {
+    width: 99%;
+    height: 311px;
```
vapaj, line 9: Probably needs more height as, at least for me, the iframe doesn't fit in `311px`.
Also, should we remove borders from this (`border: 0`)?

packages/components/src/Modal.purs
```diff
@@ -0,0 +1,29 @@
+module KSF.Modal where
```
vapaj, line 1: I guess we aren't using this anywhere?

ae-mo: Yes I was already thinking I could remove it from the repo and keep it only locally, should we need to have it later.

ae-mo:
Some work has been done:

- add logging

- add Swedish text

- remove unused imports

This is what's left:

- I need to make a better job hiding the loading of a user's credit cards

- I need to hide the ugly nets iframe during processing behind a loader

- Implement the credit card choice menu

@vapaj should I take the time to do at least part of what's left or can we do it later?

less/CreditCard.less
ae-mo: I tried removing  borders but it makes the right side disappear while the others are still there (I guess it's out of our control).

> Probably needs more height as, at least for me, the iframe doesn't fit in 311px

I wonder what might be the reason for this, because in my screen things are fine. Isn't there any relative way to specify height? 🤔  Maybe not since we don't control the content of the iframe.

vapaj: > Maybe not since we don't control the content of the iframe

This is indeed the case

vapaj: In vetrina we have the following styles for the iframe:
```
.vetrina--payment-terminal {
    background: #fff;
    width: 550px;
    height: 360px;
    border: none;
}
```

ae-mo: Ok let's think about this in another issue, since I can't seem to find a nice way to display the iframe in the account edit box.

vapaj (approved):

PR #463, opened 2020-11-09
Add IconAction component

vapaj:
Adds `IconAction` component. We are using the same style of `<icon> <description>` links/actions in a couple of places in mitt konto (and underlying components) so this will make it a bit nicer. This also makes a clear distinction of css classes and module responsibilities regarding these actions.

### DIFF
--- a/apps/mitt-konto/src/MittKonto/AccountEdit.purs
+++ b/apps/mitt-konto/src/MittKonto/AccountEdit.purs
@@ -3,6 +3,7 @@ module MittKonto.AccountEdit where
 import Prelude
 
 import Bottega.Models (CreditCard)
+import Data.Array (cons)
 import Data.Array as Array
 import Data.Either (Either(..))
 import Data.Maybe (Maybe(..))
@@ -16,7 +17,10 @@ import KSF.User as User
 import MittKonto.ActionsWrapper as ActionsWrapper
 import React.Basic (make, JSX)
 import React.Basic as React
+import React.Basic.DOM as DOM
 import React.Basic.Events (EventHandler, handler_)
+
+import KSF.IconAction as Aa
 
 type Self = React.Self Props State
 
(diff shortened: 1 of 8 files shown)


ae-mo:
This might be a good occasion to do a bigger refactoring: move `Subscription`, `TemporaryAddressChange`, `PauseSubscription` and `DeliveryReclamation` in the `mitt-konto` app. 
Moreover, the `ActionsWrapper` was created with the intention of using it for the subscription update actions as well, so this is probably a good moment to do this too.

vapaj:
> I mean, a different PR could do it too

Sounds like a few PRs already! 😄 As a note, I'm fine having these components under `packages/components` even though they aren't used, and probably will not ever be used, outside of `apps/mitt-konto`. However, these components do not have any dependencies (at least the shouldn't) for mitt konto currently, like css classes. So in that way, they are simple "stand-alone" components. What is your motivation to move them under mitt konto?

ae-mo:
> What is your motivation to move them under mitt konto?

They are there for modularizing the **mitt-konto** interface, they are not going to be reused anywhere else, unless a big twist happens. They were developed specifically to be displayed in **mitt-konto**, thus they are effectively part of it.

vapaj:
I don't disagree with you, I'm just fine separating them under `components` 😃 Having them under `mitt-konto` would probably make things clearer though.

However, if we do this, then I suggest we move everything only used by mitt konto under `mitt-konto` to be consistent. Also in that scenario I'd like the css classes to be renamed and possibly moved into `mitt-konto.less` (in any case, having the `<ComponentName>.less` doesn't fit well anymore).

vapaj:
To be fair, vetrina is structured like this and it wouldn't make sense to separate `Vetrina.Products` under `components`.

ae-mo:
> However, if we do this, then I suggest we move everything only used by mitt konto under mitt-konto to be consistent. Also in that scenario I'd like the css classes to be renamed and possibly moved into mitt-konto.less (in any case, having the <ComponentName>.less convention doesn't fit well anymore).

I agree. I guess I'll make an umbrella issue for this migration.

ae-mo (approved):

PR #466, opened 2020-11-12
Use router with payments

### LINKED ISSUE #450: Use React router to open /fakturor
Let's use React router instead of an anchor tag to open the `/fakturor` view here https://github.com/KSF-Media/affresco/blob/master/apps/mitt-konto/src/MittKonto/Main.purs#L338-L343

Something like in the Prenumerera POC: https://github.com/KSF-Media/affresco/blob/master/apps/prenumerera/src/Prenumerera/ProductSelect.purs#L169-L173

vapaj:
Adds react router to `/fakturor` path. Also prevents loading payments and credit cards of user multiple times.

Fix #450

### DIFF
--- a/apps/mitt-konto/src/MittKonto/AccountEdit.purs
+++ b/apps/mitt-konto/src/MittKonto/AccountEdit.purs
@@ -79,12 +79,12 @@ render self =
       [ IconAction.iconAction
           { iconClassName: passwordChangeClass
           , description: "Byt lösenord"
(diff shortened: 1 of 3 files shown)


kaol (approved):
I found the change to `paymentsLoad` a bit confusing. It calls `setState` but so does `didMount` in PaymentAccordion which uses the `Aff` action. It feels like it's duplicating things but I don't know right away how to avoid it either.

PR #467, opened 2020-11-13
Fix bad routes

vapaj:
Render mitt konto whenever we encouter a route we don't recognize


apps/mitt-konto/src/MittKonto/Main.purs
```diff
@@ -103,25 +103,26 @@ render self@{ state, setState } =
...
          { exact: true
          , path: toNullable $ Just "/"
-         , render:
-             \_ -> classy DOM.div "mitt-konto--container clearfix"
-               [ foldMap loadingIndicator state.loading
-               , case state.loggedInUser of
-                   Just user -> userView self user
-                   Nothing   -> loginView self
-               ]
+         , render: \_ -> mittKontoView
```
ae-mo, line 116: I guess we could use `const` here as well?

apps/mitt-konto/src/MittKonto/Main.purs
```diff
@@ -134,6 +135,13 @@ render self@{ state, setState } =
                    Nothing   -> loginView self
                ]
          }
+   noMatchRoute =
+      element
+        Router.route
+          { exact: false
+          , path: toNullable Nothing
```
ae-mo, line 142: Just asking for lack of knowledge: why here we pass `Nothing`? What happens if we pass `Just "/"`

vapaj: Then it would match the route `"/"` (which we already handle), but we want it to match everything if all other routes fail.

The router will choose the first one to match from this array of routes:
```
Router.switch { children: [ paymentListRoute, mittKontoRoute, noMatchRoute ]
```
By passing `path: null` to the `Route` component, it matches everything as it says in the documentation: https://reactrouter.com/web/api/Route/path-string-string

apps/mitt-konto/src/MittKonto/Main.purs
vapaj: Yes

apps/mitt-konto/src/MittKonto/Main.purs
vapaj: We should probably use `Redirect` here (to root), although it's not yet implemented in our router wrapper 😄

ae-mo (approved):

PR #470, opened 2020-11-20
Bump packages.dhall to psc-0.13.8-20201021 to bring purescript-react-…

kaol:
…basic 15.0.0

Most of the needed code changes involve changing the import of `React.Basic` to `React.Basic.Classic`.

### DIFF
--- a/apps/mitt-konto/src/MittKonto/Main.purs
+++ b/apps/mitt-konto/src/MittKonto/Main.purs
@@ -35,8 +35,9 @@ import KSF.Subscription.Component (subscription) as Subscription
 import KSF.User (User, UserError(..), SubscriptionPayments)
 import KSF.User (logout, getPayments) as User
 import KSF.User.Login (login) as Login
-import React.Basic (JSX, element, make)
-import React.Basic as React
+import React.Basic (JSX)
+import React.Basic.Classic (element, make)
+import React.Basic.Classic as React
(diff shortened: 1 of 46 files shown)


ae-mo (approved):

apps/mitt-konto/src/MittKonto/Main.purs
```diff
@@ -72,37 +68,36 @@ setPayments payments = _ { payments = payments }
...
+    state /\ setState <- useState initialState
+    let self = { state, setState }
+    pure $ render self logger
+
+jsApp :: {} -> JSX
+jsApp = unsafePerformEffect app
 
-render :: Self -> JSX
-render self@{ state, setState } =
+render :: Self -> Sentry.Logger -> JSX
```
ae-mo, line 97: Is it unavoidable to pass the logger all the time?

ae-mo: It's not a real problem, I was just wondering if it was forced by **react**

kaol: I could have left it just as it was, but my idea was that it's better this way since the only reason we used `setState` to set the logger was because we used `didMount` to set it. No actions done after that point ever changed it. The fewer things we have as state the better.

ae-mo (approved):

PR #475, opened 2020-11-30
User search and personating interface for admin users

kaol:
Adds a new type, Credentials, which is used for auth credential
storage instead of reusing LoginResponse which is used by Persona.

### DIFF
--- a/apps/mitt-konto/src/MittKonto/Main.purs
+++ b/apps/mitt-konto/src/MittKonto/Main.purs
@@ -2,20 +2,26 @@ module MittKonto.Main where
 
 import Prelude
 
+import Data.Either (Either(..))
 import Data.Foldable (foldMap)
 import Data.Maybe (Maybe(..))
 import Data.Nullable (toNullable)
 import Effect (Effect)
+import Effect.Aff as Aff
 import Effect.Unsafe (unsafePerformEffect)
 import MittKonto.Main.Elements as Elements
 import MittKonto.Main.Helpers as Helpers
 import MittKonto.Main.Types as Types
 import MittKonto.Main.Views (alertView, footerView, loginView, navbarView, paymentView, userView) as Views
+import KSF.Alert.Component as Alert
 import KSF.Paper (Paper(..))
+import KSF.Search as Search
 import KSF.Sentry as Sentry
+import KSF.Spinner as Spinner
+import KSF.User as User
 import React.Basic (JSX, element)
 import React.Basic.DOM as DOM
-import React.Basic.Hooks (Component, component, useState, (/\))
+import React.Basic.Hooks (Component, component, useState, useState', (/\))
 import React.Basic.Hooks as React
 import React.Basic.Router as Router
 
@@ -25,9 +31,11 @@ app :: Component {}
 app = do
   sentryDsn <- sentryDsn_
   logger <- Sentry.mkLogger sentryDsn Nothing "mitt-konto"
+  search <- Search.search
   let initialState =
         { paper: KSF
-        , loggedInUser: Nothing
+        , adminMode: false
+        , activeUser: Nothing
         , loading: Nothing
         , showWelcome: true
         , alert: Nothing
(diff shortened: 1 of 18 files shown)


packages/components/src/Persona.purs
```diff
@@ -559,3 +569,13 @@ getPayments uuid token =
...
              }
+
+type Forbidden = PersonaError
+  ( forbidden :: { description :: String } )
+
+searchUsers :: String -> Credentials -> Aff (Array User)
+searchUsers query auth = do
+  callApi adminApi "adminSearchGet" [ unsafeToForeign query ]
+    { authorization: oauthToken auth.token
+    , authUser: auth.uuid
```
vapaj, line 580: This will end up being the wrong uuid if there are users searched and set to local storage.

Don't know if I like this strategy of using the `uuid` of local storage to serve different purposes. It was designed to be one part of the authentication tuple of (user uuid, auth token), but now it can be something very different too. I suspect this might yield weird bugs in the future.

Didn't still think this all the way through, but if local storage is required, could we leave the original uuid and token untouched and add an optional `activeUserUuid` to indicate if we want to impersonate a customer?

vapaj:
I think we could simplify this a bit, although maybe I'm unable to consider everything here.

What if instead of having [activeUser](https://github.com/KSF-Media/affresco/pull/475/files#diff-ecb6bd3764c3dce2ace6a968d66efbda3b3612c58bf0ab785a8f9f2815019986R51), we'd have `impersonatingUser` (there's maybe a small semantic difference that might clarify my train of thought)? Without touching any other logic regarding the local storage, authenticating to Persona or the `loggedInUser`, we would just set the `impersonatingUser` in the [Search.props.setActiveUser](https://github.com/KSF-Media/affresco/pull/475/files#diff-5deb85cd2f3c6cbcc650ee581a21aacfdd348877c9039ff18a2686f5f12ea98cR30). Then, as the only place(?) we need the possible user we're impersonating is the `userView :: Self -> Sentry.Logger -> User -> JSX`, we could simply pass it whichever user we have at hand:
```
case state.impersonatingUser <|> state.loggedInUser of
  Just user -> userView self logger user
  Nothing   -> loginView self logger
```
As we have added the `isAdmin` value to the login response of Persona, we need to add that to the local storage alongside the other authentication data.

The type 
```
type Credentials =
  { token     :: Token
  , uuid      :: UUID
  , adminUuid :: Maybe UUID
  }
```
isn't really about credentials/authentication if the `uuid` and `adminUuid` are not the same (for which does the `token` belong to?). I think we can drop the `adminUuid` from this, as it should not matter if the `Credentials` belong to an admin or not. Some functions in the `Persona` module needs to be updated, as many of them are using the uuid found in local storage. Meaning, the functions should always take the `UUID` and the `Credentials` separately. However this more or less a separate issue (you need it here of course for getting the user) that is related to the admin capabilities of Persona.

packages/components/src/Persona.purs
kaol: I made first page load to always reset the uuid to the adminUuid if it is set, so that acts as a fail safe to reset the uuid to the one of the logged in user.

A pattern I'm seeing here is that uuid and token are common parameters that are passed with just about every call and uuid is the one that may need to vary based on search, but never between any sequence of calls. I was already thinking of bringing `ReaderT` to it but my changes seemed to be large enough even as is and I bypassed it by using `LocalStorage` as an ad hoc State monad. But I guess it's better to add `ReaderT` to this after all, even if it would take a bit longer to implement.

packages/user/src/User/User.purs
```diff
@@ -157,16 +160,19 @@ getUser maybeInvalidateCache uuid token = do
       Console.info "User fetched successfully"
       pure user
 
+isAdminUser :: Effect Boolean
+isAdminUser = isJust <$> LocalStorage.getItem "adminUuid"
```
vapaj, line 164: Should this be `isAdmin` ?

kaol: Yes it should, that was left over from the previous version. Will fix.

apps/mitt-konto/src/MittKonto/Main.purs
```diff
@@ -88,21 +92,40 @@ app = do
...
+        -- The user data in the search results isn't quite complete.
+        -- We do another fetch to get it all.
+        setActive result = case result of
+          Left _ -> setState $ setAlert $ Just
+                      { level: Alert.warning
+                      , title: "Laddningen misslyckades."
+                      , message: "Något gick fel."
+                      }
+          Right user -> do
+            Console.log $ "setstate active user " <> user.cusno
```
vapaj, line 106: I suppose this is only for dev debugging purposes so let's remove it 🙂

apps/mitt-konto/src/MittKonto/Main.purs
```diff
@@ -88,21 +92,40 @@ app = do
...
+            setState $ setActiveUser $ Just user
+            setPersonating true
```
vapaj, line 108: Just a minor observation: this will stay `true` even when I go back to my own account through search. Not a big deal and don't know if needs to be fixed. The only way to "reset" mitt konto is to refresh the page though.

kaol: True, this was one thing I was hoping to have the new router for.

packages/components/src/Persona.purs
```diff
@@ -559,3 +554,12 @@ getPayments uuid token =
...
              , discount: x.discount
              }
+
+type Forbidden = PersonaError
+  ( forbidden :: { description :: String } )
+
+-- Pass dummy uuid to force authUser field generation.
+searchUsers :: String -> UserAuth -> Aff (Array User)
+searchUsers query auth = do
+  callApi adminApi "adminSearchGet" [ unsafeToForeign query ]
```
vapaj, line 564: This will need to use `/users/search`, we'll get the new clients out shortly!

vapaj (approved):

PR #476, opened 2020-12-02
Retry login after a short delay if Persona returned a 503

kaol:


packages/user/src/User/Login.purs
```diff
@@ -198,38 +199,61 @@ failOnEmailMismatch self email
...
+      handleError err
+      pure true
+    withRetry retry = do
+      user <- action
+      notRetry <- case user of
+        Right user' -> do
+          handleSuccess user'
+          pure true
+        Left ServiceUnavailable -> do
+          if retry then handleError' ServiceUnavailable else do
```
vapaj, line 215: So if `retry = true` we don't actually retry? 😁 My brain is having hard time accepting this 😂

packages/user/src/User/Login.purs
```diff
@@ -198,38 +199,61 @@ failOnEmailMismatch self email
...
+            delay $ Milliseconds 1000.0
+            withRetry true
+            pure false
+        Left err -> handleError' err
+      when notRetry $ handleFinish user
```
vapaj, line 220: I think this would be neater to read without `notRetry`, even though it becomes a bit repetitive:
```
    withRetry retry = do
      user <- action
      case user of
        Right user' -> do
          handleSuccess user'
          handleFinish user
        Left ServiceUnavailable -> do
          if retry
          then do
            handleError ServiceUnavailable
            handleFinish user
          else do
            delay $ Milliseconds 1000.0
            withRetry true
        Left err -> do
          handleError err
          handleFinish user
```

packages/user/src/User/Login.purs
kaol: I thought about it the way that `retry = true` if we are already doing the retry.

I was also considering using an integer but went with a boolean since I used only one retry. Perhaps it could be clearer to use it after all.

packages/user/src/User/Login.purs
kaol: I updated the function. I moved `handleFinish` from inside the `withRetry` to a bind immediately after calling `withRetry` on the main `loginWithRetry` function. I think this emphasizes it that no matter which way you run the function, that's what gets used last.

vapaj (approved):

Issue #481, opened 2020-12-16
New Frontpage


ae-mo:
We should implement a new frontpage for **mitt konto**. The picture of the final result is shown below:

![image](https://user-images.githubusercontent.com/5704182/102321417-5211d500-3f86-11eb-8e65-4044786bf4e1.png)

It should be possible to display it both in Swedish and Finnish.

PR #482, opened 2020-12-16
Refactor MittKonto.Main

ae-mo:
I kind of went berserk on this one, but I really thought the `MittKonto.Main` module needed to be decomposed.

I'm appointing many people as reviewers because this has a big impact so I think it should be discusses and eventually adapted to accomodate everyone's point of view.

The idea that guided me is to try to identify what never changes and bundle it together in various thematic modules, as well as trying to separate what is likely to change more or less independently from the rest.

I hope this will ease our development on **mitt konto**. What motivated this in particular is the need for a new frontpage (#481 ), which needs to support the Finnish language too. I was afraid `MittKonto.Main` would have become a total mess if not decomposed.

### DIFF
--- a/apps/mitt-konto/src/MittKonto/Main.purs
+++ b/apps/mitt-konto/src/MittKonto/Main.purs
@@ -20,6 +20,9 @@ import Effect.Exception (Error, error, message)
 import Effect.Unsafe (unsafePerformEffect)
 import MittKonto.AccountEdit as AccountEdit
 import MittKonto.IconAction as IconAction
+import MittKonto.Main.LoginView as LoginView
+import MittKonto.Main.Helpers as Helpers
+import MittKonto.Main.Types as Types
 import KSF.Alert.Component (Alert)
 import KSF.Alert.Component as Alert
 import KSF.Api.Subscription (isSubscriptionCanceled) as Subscription
@@ -47,14 +50,14 @@ foreign import sentryDsn_ :: Effect String
 type State =
   { paper :: Paper
   , loggedInUser :: Maybe User
-  , loading :: Maybe Loading
+  , loading :: Maybe Types.Loading
   , showWelcome :: Boolean
   , alert :: Maybe Alert
   , payments :: Maybe (Array SubscriptionPayments)
   , creditCards :: Array CreditCard
   }
 
-setLoading :: Maybe Loading -> State -> State
+setLoading :: Maybe Types.Loading -> State -> State
 setLoading loading = _ { loading = loading }
 
 setLoggedInUser :: Maybe User -> State -> State
@@ -65,8 +68,6 @@ setAlert alert = _ { alert = alert }
 
 setPayments :: Maybe (Array SubscriptionPayments) -> State -> State
 setPayments payments = _ { payments = payments }
-
-data Loading = Loading
 
 type Self =
   { state :: State
@@ -98,10 +99,10 @@ render :: Self -> Sentry.Logger -> JSX
 render self@{ state, setState } logger =
   React.fragment
     [ navbarView self logger
-    , classy DOM.div "mt4 mb4"
+    , Helpers.classy DOM.div "mt4 mb4"
         [ foldMap alertView state.alert ]
-    , classy DOM.div "mt4 mb4 clearfix"
-        [ classy DOM.div "mitt-konto--main-container col-10 lg-col-7 mx-auto"
+    , Helpers.classy DOM.div "mt4 mb4 clearfix"
+        [ Helpers.classy DOM.div "mitt-konto--main-container col-10 lg-col-7 mx-auto"
             [ element Router.switch { children: [ paymentListRoute, mittKontoRoute, noMatchRoute ] } ]
         ]
     , footerView
@@ -115,11 +116,11 @@ render self@{ state, setState } logger =
          , render: const mittKontoView
          }
    mittKontoView =
-      classy DOM.div "mitt-konto--container clearfix"
+      Helpers.classy DOM.div "mitt-konto--container clearfix"
         [ foldMap loadingIndicator state.loading
         , case state.loggedInUser of
             Just user -> userView self logger user
-            Nothing   -> loginView self logger
+            Nothing   -> LoginView.loginView self logger
         ]
    paymentListRoute =
      element
@@ -127,11 +128,11 @@ render self@{ state, setState } logger =
          { exact: true
          , path: toNullable $ Just "/fakturor"
          , render: const
-             $ classy DOM.div "mitt-konto--container clearfix"
+             $ Helpers.classy DOM.div "mitt-konto--container clearfix"
                  [ foldMap loadingIndicator state.loading
                  , case state.loggedInUser of
                      Just user -> paymentView self user
-                     Nothing   -> loginView self logger
+                     Nothing   -> LoginView.loginView self logger
                  ]
          }
    noMatchRoute =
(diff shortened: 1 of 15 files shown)


apps/mitt-konto/src/MittKonto/Main/Views/LoginView.purs
```diff
@@ -0,0 +1,99 @@
+module MittKonto.Main.Views.LoginView where
```
kaol, line 1: I would prefer these without the "View" suffix. They're all views in that directory.

apps/mitt-konto/src/MittKonto/Main/Helpers.purs
```diff
@@ -0,0 +1,93 @@
...
+  -> String
+  -> (Array JSX -> JSX)
+classy element className children = element { className, children }
+
+anchor :: String -> String -> Array JSX -> JSX
+anchor href description children = DOM.a { href, children: DOM.text description : children, target: "_blank" }
+
+-- | Allows to run the asynchronous action while showing the loading indicator
+--   and handling the result.
+withSpinner :: forall a. (Maybe Types.Loading -> Effect Unit) -> Aff a -> Aff a
```
vapaj, line 65: We could use [KSF.Spinner](https://github.com/KSF-Media/affresco/blob/master/packages/components/src/Spinner.purs) for this (I suppose it's the same code 😅). Same goes for the `Loading` type

apps/mitt-konto/src/MittKonto/Main/Elements.purs
```diff
@@ -0,0 +1,28 @@
...
+disappearingBreak =
+  DOM.div
+    { className: "mitt-konto--disappearing-break"
+    , children: [ break ]
+    }
+
+loadingIndicator :: Types.Loading -> JSX
+loadingIndicator Types.Loading =
+  DOM.div
+      { className: "mitt-konto--loading flex items-center"
```
kaol, line 21: Could you please indent these to match `disappearingBreak`?

apps/mitt-konto/src/MittKonto/Main/Helpers.purs
ae-mo: Ah yes I was thinking about this too 👌

apps/mitt-konto/src/MittKonto/Main/Views/UserView.purs
```diff
@@ -0,0 +1,151 @@
...
+    componentBlock headerText content =
+      DOM.div
+        { className: "mitt-konto--component-block-container"
+        , children:
+            componentHeader headerText
+            : content
+        }
+
+    componentBlockContent child =
+       DOM.div
```
kaol, line 148: Align with `componentBlock

apps/mitt-konto/src/MittKonto/Main.purs
```diff
@@ -143,347 +100,24 @@ render self@{ state, setState } logger =
...
-     Aff.killFiber (error "Timeout reached") loadingFiber
-   Aff.joinFiber actionFiber # Aff.finally do
-     -- finally in the end, when the action has been completed
-     -- we kill all other threads and switch the loading off
-     Aff.killFiber (error "Action is done") timeoutFiber
-     Aff.killFiber (error "Action is done") loadingFiber
-
 -- | Navbar with logo, contact info, logout button, language switch, etc.
-navbarView :: Self -> Sentry.Logger -> JSX
+navbarView :: Types.Self -> Sentry.Logger -> JSX
```
kaol, line 104: How about moving the rest to the Views directory as well? Clump the minor ones into a Misc.purs or something.

apps/mitt-konto/src/MittKonto/Main/Views/LoginView.purs
ae-mo: 🤔  I have to confess I prefer having the module name mirroring the path of the file in general, I think it's clearer. If I was forced to name one positive aspect of Java, that's what I would mention 😆 

But I see your point! Let's change it

apps/mitt-konto/src/MittKonto/Main.purs
```diff
@@ -1,78 +1,35 @@
 module MittKonto.Main where
 
-import Prelude
+import Prelude (bind, const, discard, pure, when, ($), (<<<))
```
vapaj, line 3: I think `Prelude` should be imported fully

apps/mitt-konto/src/MittKonto/Main.purs
vapaj: How about instead of Misc.purs, move them to the main Views.purs directly?

ae-mo: I think I prefer Valtteri's suggestion! But In any case I was bothered too by having these remaining views in `Main.purs`.

> Modules named like Misc or Helpers or something gives me the creeps for some reason 🙀 They are usually some sort of dumpsters

Yes I agree in general, but I have to say that in this case I couldn't find a better alternative. I don't know sometimes I think that how **React** is made makes it difficult to modularize code 🤔

apps/mitt-konto/src/MittKonto/Main.purs
f-f: The PS compiler expects only one import to be unspecified (which is usually the Prelude), so rather than specifying that one I would specify any other open imports there are.

apps/mitt-konto/src/MittKonto/Main.purs
ae-mo: > Yes I agree in general, but I have to say that in this case I couldn't find a better alternative.

I was talking about the `Helpers` module there

apps/mitt-konto/src/MittKonto/Main.purs
ae-mo: Ah ok then it must be because I was fixing imports all at once, but good to know!

apps/mitt-konto/src/MittKonto/Main/Helpers.purs
```diff
@@ -0,0 +1,62 @@
...
+import KSF.Alert.Component (Alert)
+import KSF.Alert.Component as Alert
+import KSF.Error as KSF.Error
+import KSF.Spinner as Spinner
+import KSF.User (SubscriptionPayments, User)
+import MittKonto.Main.Types as Types
+import React.Basic (JSX)
+import React.Basic.DOM as DOM
+
+setLoading :: Maybe Spinner.Loading -> Types.State -> Types.State
```
kaol, line 24: I feel these would fit better within the same file where the type they're manipulating is located in. They seem pretty lonely in here and checking on what they do would require opening that other file as well.

ae-mo: Most of them are not used in a single file anymore and I'm strongly against duplication. Moreover I think that their names are self explanatory, I think I would only check them if I had to change them and that's a big part of the point of this refactoring to me.

Thoughts?

kaol: I meant moving them to the same file where State is defined, not copying them.

ae-mo: Aaah so in the `Types.purs` file! Yes that is a very good idea, sorry I really misunderstood :smile: 

So you mean the `setSomething` functions right?

apps/mitt-konto/src/MittKonto/Main.purs
vapaj: Yeah, it's fine 🙂

vapaj (approved):

PR #484, opened 2020-12-22
Payment detail page

kaol:


kaol:
I gave it a stab to implement this using RouterWrapper but it was a bit uneasy match. Unless I come up with a better idea, I'll leave this as is.

apps/mitt-konto/src/MittKonto/Main.purs
```diff
@@ -117,22 +127,6 @@ render self@{ state, setState } logger searchView isPersonating =
...
-       }
    noMatchRoute =
      -- TODO: Use Redirect when supported!
-     Router.route
-       { exact: false
-       , path: Nothing
-       , render: const mittKontoView
-       }
\ No newline at end of file
+     routeElement false true Nothing $ Views.userView self logger
```
ae-mo, line 132: Does this solve the problem that the @vapaj was talking about? I.e. unsopported routes don't get redirected to home.

kaol: Not, it doesn't. I just wanted to refactor the routes list a bit as the where clauses had got pretty repetitive.

ae-mo (approved):

PR #491, opened 2021-01-12
Select first day for permanent address change

kaol:


ae-mo:
I've tried to make two address changes, one taking effect today (`14/01/2020`) and another one taking effect the 28th of January 2021.

I don't think the two changes are interfering with each other, it seems the first took effect immediately. But now this is what I see:
localhost_8000_ (13)

If I'm interpreting correctly, it says the change is valid from `14/01`? Shouldn't it be from `28/01` as I specified when filling the form? Or is it displaying the date the address change was requested? If so, the date it takes effect should be there too I think.

ae-mo (approved):

PR #493, opened 2021-01-14
Client side range checks for subscription pauses

kaol:


ae-mo:
I have tried it locally and I am a bit concerned about how it behaves. Here's what I did:

1. I have created a pause for my subscription from 28/01/2010 to 03/02/2021
2. I have tried to create another pause starting from the first date that wouldn't trigger the warning: localhost_8000_ (14)

The operation did not succeed as shown below:
localhost_8000_ (15)

Now, as a user I would probably expect that when the warning is gone the operation will work, as the warning message seems to account for the fact that there should be at least one week between two pauses.

Since we're on it, should we account for this when we check that the selected range is correct?

packages/components/src/PauseSubscription/Component.purs
```diff
@@ -143,13 +148,28 @@ render self =
...
+                          , maxEndDate = maxEndDate
+                          }
+
+    invalidDateWarning =
+      DOM.div
+        { className: "subscription--error-container"
+        , children:
+            [ DOM.div
+                { className: "error-text"
+                , children: [ DOM.text "Vänligen välj ett annat datum. Det valda datumet går över eller är för nära ett tidigare uppehåll. Det måste vara minst en vecka mellan uppehållsperioderna." ]
```
ae-mo, line 170: @pikkuruusu did you provide this text? Isn't it a bit too imperative?

ae-mo:
@kaol what about this PR?

PR #494, opened 2021-01-14
Show subscription receiver name

kaol:


packages/components/src/Subscription/Component.purs
```diff
@@ -134,6 +135,11 @@ render self@{ props: props@{ subscription: sub@{ package } } } =
        else subscriptionUpdates)
       $ Just { extraClasses: [ "subscription--container" ] }
   where
+    receiverName =
+      maybe mempty (\receiver -> Array.singleton
```
vapaj, line 139: `foldMap` would fit here nicely

vapaj (approved):

PR #497, opened 2021-01-15
Vetrina/campaign page mode

vapaj:
This is huge, sorry. Changes in a nutshell:
- adds `minimalLayout` prop, which if true, will remove most of the texts and product descriptions (this is for the campaign pages mainly)
- if multiple products given, show them in a dropdown
- adds `paymentMethods` prop
  - if no payment methods given, credit card is chosen by default
  - if multiple payment methods given, a radio button group is shown
- adds support for paper invoices
- `Product` takes now a full `Campaign` as a prop instead of just the campaign id
  - if any of the products given has a insufficient campaign prop, e.g. with missing fields, error will be thrown (the idea is to make this clearer to non-technical people to use)

The layout is a bit crude, however it's a start. Also I'm not that proud of the whole `minimalLayout` thing as it's a step towards the codebase being a huge spaghetti mess. I propose we rethink the use cases of vetrina and redesign the whole thing.


kaol:
I would suggest using `guard (not self.props.minimalLayout) a` instead of
```
if self.props.minimalLayout
then mempty
else a
```

kaol (approved):

Issue #279, opened 2020-03-10
Use Data.UUID everywhere


vapaj:
We could use the `UUID` type from [Data.UUID](https://pursuit.purescript.org/packages/purescript-uuid/6.0.1/docs/Data.UUID) in our code (as we are already using that module for generating `UUID`'s), instead of our own type.

PR #501, opened 2021-01-18
Search: Allow directly selecting user by UUID

kaol:


packages/components/src/Api.purs
```diff
@@ -10,6 +13,13 @@ derive newtype instance showUUID :: Show UUID
 derive newtype instance readforeignUUID :: ReadForeign UUID
 derive newtype instance writeforeignUUID :: WriteForeign UUID
 
+toUUID :: String -> Maybe UUID
+toUUID uuid = if test regex uuid then Just $ UUID uuid else Nothing
+  where
+    test (Left _) = const false
+    test (Right r) = Regex.test r
+    regex = Regex.regex "^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$" mempty
```
ae-mo, line 21: 🤔  can this part of the regex: 

```
[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-
```

Be compacted? Something like:

```
[[0-9a-f]{4}-]{3}
```

Don't know if this is correct syntax, but hopefully you get the point 😆

vapaj: We could actually use `Data.UUID` here (https://pursuit.purescript.org/packages/purescript-uuid/7.0.0/docs/Data.UUID), as this function is implemented there already.

FYI: There's also an issue to replace our `UUID` type with `Data.UUID` (https://github.com/KSF-Media/affresco/issues/279) but we don't need to do that in this PR.

ae-mo (approved):

Issue #469, opened 2020-11-19
Move update credit card functionality to individual subscription boxes


ae-mo:
We should move the functionality to update a credit card to subscription boxes. 

This way, the functionality could be found more easily by a customer. Also the customer would be able to choose an existing credit card to replace the old one, as well as managing her credit cards (deleting unused ones).

ae-mo:
I think `MittKonto.Main` could be modularized as well. Currently I find it quite hard to navigate it, so maybe we are not taking full advantage of **react** modules.

ae-mo:
Closing for now, as we have a basic version of this and there are other priorities.

PR #504, opened 2021-01-22
Patch/credit card update route

ae-mo:
Related to #469.

This PR changes the credit card update functionality to be displayed full screen in a separate route.

It also introduces the `RouteWrapper` component which should help us dealing with displaying stuff full screen.


apps/mitt-konto/src/MittKonto/Main/Views/CreditCardUpdateView.purs
```diff
@@ -0,0 +1,188 @@
...
+startRegisterPoller setState props state oldCreditCard creditCardRegister = do
+  newPoller <- Aff.forkAff do
+    killRegisterPoller state
+    newPoller <- Aff.forkAff $ pollRegister setState props state oldCreditCard (Right creditCardRegister)
+    Aff.joinFiber newPoller
+  liftEffect $ setState _ { poller = newPoller }
+
+pollRegister :: SetState -> Props -> State -> CreditCard -> Either BottegaError CreditCardRegister -> Aff Unit
+pollRegister setState props@{ logger, onError, onSuccess, onCancel } state oldCreditCard (Right register) = do
+  Aff.delay $ Aff.Milliseconds 1000.0
```
kaol, line 163: This doesn't feel like the right place for waiting. We know what the value of `register.status.state` is even before calling `delay`, why not act on it right away? If the value is `CreditCardRegisterCompleted` then won't this just add one more second to the whole process?

apps/mitt-konto/src/MittKonto/Wrappers/ViewWrapper.purs
```diff
@@ -0,0 +1,164 @@
...
+  where
+    header :: JSX
+    header = Grid.row_
+      [ DOM.div
+          { className: "col col-11"
+          , children: [ title ]
+          }
+      , if closeable then
+          DOM.div
+            { className: "col-1 flex credit-card-choice--close-icon"
```
kaol, line 140: I take it that `ViewWrapper.purs` is meant for other users besides credit cards? This class name wouldn't belong here then.

packages/components/src/Navbar/Component.purs
```diff
@@ -12,10 +12,9 @@ import Effect (Effect)
 import KSF.Icons (papers)
```
kaol, line 12: I would leave pure style changes to a different pull request, this is large enough even without and they risk merge conflicts.

apps/mitt-konto/src/MittKonto/Wrappers/ViewWrapper.purs
ae-mo: Yes of course that is not right 😃

packages/components/src/Navbar/Component.purs
ae-mo: 🤔  I don't think I have anything to do with this 😅

apps/mitt-konto/src/MittKonto/Main/Views/CreditCardUpdateView.purs
ae-mo: It's a trick to poll every second, but we can think of a better place for that delay sure.

I never really questioned it because I started this component by copying **vetrina**'s code, but it is certainly good to scrutinize it, cause that delay is indeed annoying when the register operation has succeded.

ae-mo:
The route of the credit card update view is currently `/kortt/uppdatera`, but I think there might be a better alternative 😅 I don't even know if it's correct Swedish

vapaj:
> The route of the credit card update view is currently `/kortt/uppdatera`, but I think there might be a better alternative  I don't even know if it's correct Swedish

I think `/kort/uppdatera` (with one t) is close enough 😅 Or even `/kreditkort/uppdatera` to be precise.

packages/components/src/Navbar/Component.purs
ae-mo: @kaol What did you exactly mean? I don't know how to act on this :blush:

kaol: Never mind, now that I looked a bit closer I saw that this was needed since `User` was added to User.purs. I first thought that this just rearranged imports to no effect.

apps/mitt-konto/src/MittKonto/Wrappers/RouteWrapper.purs
```diff
@@ -0,0 +1,126 @@
...
+            { className: "route-wrapper"
+            , children:
+                [ header
+                , renderedContent
+                ] <> [ autoClose self.props closeAutomatically ]
+            }
+        }
+      where
+        header :: JSX
+        header = Grid.row_
```
vapaj, line 73: I would not like to use the `Grid` module we have anymore in new code as it's dependent on [basscss](https://basscss.com/) which we don't need IMO. I think it's better to use plain css grids as they are really straight forward.

vapaj:
Don't know where to exactly leave this comment to, but the `noMatchRoute` does not work anymore in this branch 🤔 So for instance `http://localhost:8000/asdasdasds` should render the mitt konto front page but it doesn't

apps/mitt-konto/src/MittKonto/Main/Views/CreditCardUpdateView.purs
```diff
@@ -0,0 +1,194 @@
...
+didMount self@{ setState, props: { creditCards, logger, setWrapperState } } = do
+  setState _ { asyncWrapperState = AsyncWrapper.Ready }
+  Aff.launchAff_ do
+    case creditCards of
+      []       -> liftEffect $ do
+        logger.log "No credit cards found" Sentry.Error
+        onError self
+      [ card ] -> do
+        registerCreditCard self card
+      _        -> pure unit
```
vapaj, line 86: Should we take `Maybe CreditCard` (or do we even need `Maybe`?) in props instead of `Array CreditCard` if only one card makes sense in this component?

vapaj: Ohh, you can choose if there are multiple cards, but if there's only one, it's registered automatically?

ae-mo: > Ohh, you can choose if there are multiple cards, but if there's only one, it's registered automatically?

Yes that is the deal

ae-mo:
> Don't know where to exactly leave this comment to, but the `noMatchRoute` does not work anymore in this branch 🤔 So for instance `http://localhost:8000/asdasdasds` should render the mitt konto front page but it doesn't

It does not work in prod either, I've checked. I think this is better suited for a new issue.

kaol (approved):

PR #511, opened 2021-01-31
Fix lettera baseurl

mrskug:


mrskug:
fixing this in a more consistent way in #512

PR #517, opened 2021-02-08
Bugfix/cookie user

lfaz:
add expire when setting cookie


apps/app-article/src/App.js
```diff
@@ -460,14 +459,22 @@ if (window.ksfDfp) {
...
+        Cookies.set('uuid', localStorage.getItem('uuid'), { expires: 365 });
         //To get User data from Android side 
-        Cookies.set('currentUser', JSON.stringify({ firstName: user.firstName, lastName: user.lastName, email: user.email, token: localStorage.getItem('token'), uuid: localStorage.getItem('uuid') }));
+        Cookies.set('currentUser', JSON.stringify({ firstName: user.firstName, lastName: user.lastName, email: user.email, token: localStorage.getItem('token'), uuid: localStorage.getItem('uuid') }, {expires: 365}));
         localStorage.setItem("currentUser", JSON.stringify(user));
         this.setState({user: user});
-        this.fetchArticleFromApi(getUrlParam().has('uuid')?getUrlParam().get('uuid'):"");
-        //Call Android bridge 
-        Android.isLoggedIn();        
+        if(getUrlParam().has('uuid') && getUrlParam().has('uuid') != ""){
```
vapaj, line 469: `getUrlParam().has('uuid') != ""` compares boolean with string

apps/app-article/src/App.js
```diff
@@ -460,14 +459,22 @@ if (window.ksfDfp) {
...
+            this.fetchArticleFromApi(getUrlParam().has('uuid')?getUrlParam().get('uuid'):"");
```
vapaj, line 470: Could we assign `getUrlParam().get('uuid')` to a variable and work with that instead of checking the url params multiple times? 🙂 Something like:
```
const articleUuid = getUrlParam().get('uuid');
if (articleUuid) { 
  this.fetchArticleFromApi(articleUuid);
}
```
This would make the code a bit easier for the reader.

lfaz: yes this looks much cleaner thnx! already commited the change!

vapaj (approved):

PR #518, opened 2021-02-08
Hooks version of RouteWrapper

kaol:


ae-mo (approved):

apps/mitt-konto/src/MittKonto/Main.purs
```diff
@@ -78,13 +79,31 @@ app = do
                        }
         paymentView = payments paymentProps
         paymentDetailView = paymentDetail paymentProps
-    pure $ render self logger searchView paymentView paymentDetailView isPersonating
+        creditCardUpdateInputs =
+          CreditCardUpdateView.RouteWrapperContentInputs
+            { creditCards: fromMaybe mempty $ state.activeUser <#> _.creditCards
+            , logger: logger
+            }
+        updateCreditCardRoute =
```
ae-mo, line 87: Maybe we could refactor this bit to use `defaultRouteElement` or `routeElement`? Sorry I don't know if it was like now before my approval, I missed it

apps/mitt-konto/src/MittKonto/Wrappers/RouteWrapper.purs
```diff
@@ -30,15 +31,14 @@ type SetRouteWrapperState = (State -> State) -> Effect Unit
...
-component :: forall p. React.Component (Props p)
-component = React.createComponent "RouteWrapper"
-
-routeWrapper :: forall p. (RouteWrapperContent p) => (Props p) -> JSX
-routeWrapper = make component
-  { initialState
-  , didMount
-  , render
-  }
+routeWrapper :: forall p t. Eq t => RouteWrapperContent p => (p -> t) -> Component (Props p)
```
ae-mo, line 34: :thinking: I don't get why we need to pass a function now?

kaol: Because RouteWrapper's rendering is done inside the `useEffect` in this version and `useEffect` is only called if its first parameter has changed value. That function is used to build the value that `useEffect` uses as its trigger condition.

kaol:
I thought about this a bit and I'll close this without merging. This can be dusted off once there are some hooks code that actually uses this, I thought I could soon after this  pull request follow up with changing the detailed payment view use it but that didn't yet lead anywhere.

PR #521, opened 2021-02-09
add new logout handler for logout

lfaz:
handle logout http req

### DIFF
--- a/apps/app-article/src/App.js
+++ b/apps/app-article/src/App.js
@@ -81,7 +81,7 @@ class App extends Component {
             logout(this.onLogout);
         }
         if(getUrlParam().has('logout')){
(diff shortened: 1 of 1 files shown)


apps/app-article/src/App.js
```diff
@@ -92,10 +95,14 @@ class App extends Component {
     componentWillUnmount() {
     }  
     onLogout() {
+        Android.onLogoutSuccess();
         console.log("Logged out successfully!");
         //Remove the current user from localstorage 
         localStorage.removeItem("currentUser");
+        localStorage.removeItem("token");
```
vapaj, line 102: I don't think we need to remove these here, as they are already wiped.

vapaj (approved):

Issue #464, opened 2020-11-10
Refactor Mitt Konto


ae-mo:
It would be beneficial to move all the components that are used exclusively by **Mitt Konto** inside its folder, as this would make reusing code inside **Mitt Konto** easier. Thus, we would probably achieve a higher degree of modularization.

One example: we could use `ActionsWrapper`, with is now used inside `AccountEdit`, for the `Subscription` component as well, without having to leak `ActionsWrapper` out of **Mitt Konto**'s namespace.

Things we need to do:

- Move `Subscription`, `PauseSubscription`, `TemporaryAddressChange` and `DeliveryReclamation` inside `apps/mitt-konto`
- Refactor these components' style so that CSS classes reflect that they are part of **Mitt Konto**. Perhaps we should move the stylesheets in **Mitt Konto's** folder as well.

ae-mo:
I think `MittKonto.Main` could be modularized as well. Currently I find it quite hard to navigate it, so maybe we are not taking full advantage of **react modules.

ae-mo:
> I think `MittKonto.Main` could be modularized as well. Currently I find it quite hard to navigate it, so maybe we are not taking full advantage of **react** components.

This was taken care of in #482.

PR #527, opened 2021-02-11
Refactor Subscription component

### LINKED ISSUE #469: Move update credit card functionality to individual subscription boxes

ae-mo:
Related to #469, related to #464. 

Before moving the update credit card button to individual subscription boxes, I figured I'd do a refactoring I wanted to do for a long time.

I moved the `Subscription` component in Mitt Konto. I also decomposed it in a similar fashion to how I did  with `MittKonto.Main`. The reason is that this component was becoming an unreadable blob, too.


kaol (changes requested):
Just some trailing whitespace cleanups and a missing colon to a string.

vapaj:
Looks good! I wish we had tests 😄

kaol (approved):

PR #528, opened 2021-02-11
Ios logout improvement, comment old cookie removal 

baderkhane:


apps/app-article/src/App.js
```diff
@@ -70,18 +70,19 @@ class App extends Component {
...
+        /*if(Cookies.get('LoginStatus') != undefined && Cookies.get('LoginStatus') === "false"){
             //we remove it to avoid infinite loop
             Cookies.remove('LoginStatus');
             //TODO
             // we need to have logout listener here, after success we can then remove the cookie, if for exemple an error happend while             
             //LogOut;in localstorage will keep logged but for android is considered as logged out
 
             logout(this.onLogout);
-        }
+        }*/
```
lfaz, line 83: maybe delete instead of commenting?

lfaz (approved):

apps/app-article/src/App.js
```diff
@@ -95,12 +84,37 @@ class App extends Component {
     componentWillUnmount() {
     }  
     onLogout() {
-        Android.onLogoutSuccess();
+        //To avoid undefined error for ios webview
+        try {
+            Android.onLogoutSuccess();
+        } catch (e) {
+            console.log(e);
```
vapaj, line 91: If the purpose is to debug, we could use `console.debug(e)` instead of `log`.

baderkhane: @vapaj Yes it can be replaced, should i send new commit ?

vapaj: I mean, the log will stay in the browser so it's not much use.

baderkhane: Actually it could be replaced with Android == 'undefined', because the reason of that try{}catch just to avoid error log that could prevent some phone version from rendering articles 
@vapaj

PR #532, opened 2021-02-16
Add messages about contacting customer service for email change

kaol:


packages/components/src/Profile/Component.purs
```diff
@@ -141,12 +141,19 @@ render self@{ props: { profile: user } } =
...
             , { term: "Kundnummer:", description: [ DOM.text user.cusno ] }
             ]
         }
     ]
   where
+    emailChangeMessage =
+      DOM.div
+        { className: "mitt-konto--note"
+        , children: [ DOM.text "Om du vill ändra din e-postadress vänligen kontakta Kundservice." ]
+        }
```
ae-mo, line 156: I've built this locally, I have a couple of comments:

1. Could the phrase 
    > vänligen kontakta Kundservice
  
    become a link to the kundservice contact page as for the login page?
2. It is a bit hard for me to decode this bit with this text immediately below the email address value. An accessibility concern if you will 🤣 maybe it could have a slightly different layout?

ae-mo (approved):

PR #535, opened 2021-02-19
Edit temporary address change dates

kaol:


kaol:
To avoid confusion, it's best if this gets merged at the same time as https://github.com/KSF-Media/affresco/pull/537 even though they're independent changes.

ae-mo:
I have encountered an error while trying to make a temporary address change, my account's and subscription's data are the same as the comment in #537.

image

ae-mo (approved):

PR #537, opened 2021-03-01
Edit subscription pause

kaol:


less/Subscription.less
```diff
@@ -138,3 +138,8 @@
     justify-content: flex-start;
     width: 100%;
 }
+
+.subscription--edit-subscription-pause {
```
vapaj, line 142: Is this used anywhere?

kaol:
Marked as WIP since I still need one string.

kaol:
To avoid confusion, it's best if this gets merged at the same time as https://github.com/KSF-Media/affresco/pull/535 even though they're independent changes.

ae-mo:
I tried to edit a sub pause, I got this error:

Screenshot 2021-03-08 at 13 16 52

My `cusno` is `9814144` if you need to check, `subsno=1017470`.

Pause dates were `10.03-11.03.2021`. I tried to change end date.

kaol:
The persona client library in package.json files was outdated, it needed an upgrade.

ae-mo (commented):
Editing a subscription by moving the start date ahead in time and using the calculated end date does not work, `persona` returns `PauseInvalidLength`:

Request:
```
PATCH /v1/users/67748eb2-7d19-42c7-9043-4fb854303cf3/subscriptions/1017470/pause
```

Request payload:
```
{"oldStartDate":"2021-03-10","oldEndDate":"2021-03-16","newStartDate":"2021-03-17","newEndDate":"2021-03-16"}
```

Response:
```
{"http_status":"Bad request","invalid_pause_dates":{"message":"PauseInvalidLength","description":"Bad request"},"http_code":400}
```

kaol:
We still don't have client side checks for subscription pauses and this pull request didn't add any, either.

kaol:
Could you please describe what was what you expected? You inputted invalid data to the backend and it got recognized as such correctly.

ae-mo:
> Could you please describe what was what you expected? You inputted invalid data to the backend and it got recognized as such correctly.

I have changed the starting date for the subscription pause and the client calculated the end date automatically, I would expect that it works since I didn't provide the end date, **Mitt Konto** did.

> We still don't have client side checks for subscription pauses and this pull request didn't add any, either.
Can we add them in this PR?

kaol:
Let's not make this any larger than what it is but I suppose we could get https://github.com/KSF-Media/affresco/pull/493 merged first and use that.

kaol:
I made the end date field not initialize with the original end date. This forces the user to pick a value and it'll follow the existing calcMinEndDate limit.

ae-mo (approved):

PR #542, opened 2021-03-02
Output cusno to console log on user fetch success

kaol:


kaol:
This pull request was a bit of a misunderstanding. It was supposed to be about `setUser` call and scopes, which got addressed in a different branch. Closing this now.

PR #553, opened 2021-03-05
Prevent pauses and temporary address changes from some packages

kaol:


packages/components/src/Api/Package.purs
```diff
@@ -96,3 +99,7 @@ type JSCampaign =
   , lengthUnit :: Nullable String
   , priceEur   :: Nullable Number
   }
+
+-- TODO: Set this as field in the API
+testPauseTemp :: Package -> Boolean
+testPauseTemp package = not (elem package.id [ "FORUM_P+D", "JUNIOR PD" ])
```
ae-mo, line 105: I think it's better to have this blacklist in one place only, instead of hardcoding it here as well.

It would be nice to have an endpoint to obtain a list of blacklisted packages.

ae-mo: But yeah setting this as an API field would be better as you say

kaol: I plan to remove this code once Persona returns information about which packages don't have these feature enabled.

kaol:
I'll do this the long way then, changes to backend first.

PR #559, opened 2021-03-08
Fix credit card update view hangs when user has more than one card

### LINKED ISSUE #558: Bug: credit card update view hangs when user has more than one card
In **Mitt Konto**, the credit card update view hangs when user has more than one card, displaying the loader until the registration process times out.

ae-mo:
Closes #558

### DIFF
--- a/apps/mitt-konto/src/MittKonto/Main/Views/CreditCardUpdateView.purs
+++ b/apps/mitt-konto/src/MittKonto/Main/Views/CreditCardUpdateView.purs
@@ -72,9 +72,8 @@ creditCardUpdateView = do
           []       -> liftEffect $ do
             logger.log "No credit cards found" Sentry.Error
             onError self
-          [ card ] -> do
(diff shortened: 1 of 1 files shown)


mfahler9:
That solves the hangup, but now it shows the cards used for each subscription, even if they are the same card, which is confusing. Maybe we could filter out duplicates?

Also, the subscriber doesn't know which card is for which subscription if they are using different credit cards.

mfahler9:
> That solves the hangup, but now it shows the cards used for each subscription, even if they are the same card, which is confusing. Maybe we could filter out duplicates?

Well, that might not be possible due to Nets. But maybe data on which card is tied to which subscription?

ae-mo:
> Maybe we could filter out duplicates?

No unfortunately

> But maybe data on which card is tied to which subscription?

We can do better, I can select the correct credit card for the subscription at hand.

mfahler9 (approved):

PR #569, opened 2021-03-15
Corona banner/fetch from api

MicciKarkia:


apps/corona-banner/src/app.js
```diff
@@ -1,52 +1,96 @@
...
+const Banner = ({
+  newCases,
+  hospitalised,
+  deaths,
+  vaccinated,
+  vaccinatedPercentage,
+}) => {
+
+  return (
+    <a href='/tagg/coronaviruset/'>
```
vapaj, line 15: This will lead to `frontends.ksfmedia.fi/corona-banner/tagg/coronaviruset` which is probably not intended 🙂

apps/corona-banner/src/app.js
```diff
@@ -1,52 +1,96 @@
...
 
   useEffect(() => {
     console.log('effect')
-    fetch("https://cat-fact.herokuapp.com/facts")
-    .then(res => console.log(res))
-    
+    axios
+      .get(' https://cdn.ksfmedia.fi/corona-banner/stats.json')
+      .then(response => {
+        console.log('promise fulfilled')
```
vapaj, line 73: Let's remove these two `console.logs`, as I suppose they are for debugging purposes (not really useful in production)

apps/corona-banner/src/index.less
```diff
@@ -0,0 +1,151 @@
+@font-face {
```
vapaj, line 1: We could move this file under the `less/` directory in the root to be consistent with other apps in the repo (e.g. mitt-konto and app-article). The naming convetion is `<app-name>.less`.

apps/corona-banner/src/app.js
```diff
@@ -1,52 +1,121 @@
...
+      <div className='source'>
+        <em>Källa: THL och Helsingin Sanomat</em>
       </div>
     </div>
-  );
+  )
+}
+
+
+  function getSiteUrl() {
```
vapaj, line 69: Let's indent this correctly (one level too many there)

vapaj (approved):

PR #576, opened 2021-03-22
Add email updating to profile component

kaol:


packages/user/src/User/User.purs
```diff
@@ -203,7 +204,11 @@ updateUser uuid update = do
   newUser <- try $ Persona.updateUser uuid update =<< requireToken
   case newUser of
     Right user -> Right <$> fromPersonaUserWithCards user
-    Left err   -> pure $ Left $ UnexpectedError err
+    Left err
+      | KSF.Error.resourceConflictError err -> do
+          Console.error "Email update unique violation"
```
vapaj, line 209: We don't probably need to log this

packages/components/src/Profile/Component.purs
```diff
@@ -379,6 +428,63 @@ editAddress self =
...
+          newUser <- User.updateUser self.props.profile.uuid $ User.UpdateEmail { email: email }
+          case newUser of
+            Right u -> liftEffect do
+              self.props.onUpdate u
+              self.setState _ { editEmail = Success Nothing }
+              Tracking.changeEmail self.props.profile.cusno "success"
+            Left UniqueViolation -> do
+              liftEffect do
+                self.setState _ { editEmail = AsyncWrapper.Error "Den här e-postadressen används för ett annat konto. Vänligen ta kontakt med kundservice om du har frågor." }
+              throwError $ error "Unique violation when updating email."
```
vapaj, line 479: Don't know if we need to throw here actually. I guess it's enough we just show the error message to the user. This would end up in Sentry though, if that was your reasoning.

packages/user/src/User/User.purs
kaol: I removed the logging.

packages/components/src/Profile/Component.purs
kaol: I removed the throw.

vapaj (approved):

PR #580, opened 2021-03-25
Add newtype Cusno in KSF.User.Cusno

kaol:


packages/components/src/Navbar/Component.purs
```diff
@@ -179,7 +180,7 @@ customerService isPersonating activeUser = do
       userLink user =
         Router.link
           { to: { pathname: "/", state: {} }
-          , children: [ DOM.text $ user.cusno <> " - " <>
+          , children: [ DOM.text $ (show $ (\(Cusno c) -> c) $ user.cusno) <> " - " <>
```
ae-mo, line 183: I think `show $ (\(Cusno c) -> c` could become a `toString` or `fromCusno` function to put in the `KSF.User.Cusno` module.

apps/mitt-konto/src/MittKonto/Main/Views/UserView/Subscription/Elements.purs
```diff
@@ -275,11 +275,11 @@ subscriptionUpdates self@{ props: props@{ subscription: sub@{ subsno, package }
                           self.setState _
                             { wrapperProgress = AsyncWrapper.Success $ Just "Tillfällig adressändring har tagits bort",
                             pendingAddressChanges = toMaybe newSubscription.pendingAddressChanges }
-                          Tracking.deleteTempAddressChange (show props.subscription.cusno) (show props.subscription.subsno) startDate' endDate' "success"
+                          Tracking.deleteTempAddressChange props.subscription.cusno (show props.subscription.subsno) startDate' endDate' "success"
```
ae-mo, line 278: I think we should add the `Subsno` type as well, but we can do it in a different PR.

ae-mo (changes requested):

apps/mitt-konto/src/MittKonto/Main/Views/UserView/Subscription/Elements.purs
kaol: I would leave it for a later one. I did this for `Cusno`s now since I needed to do something to get to compare `Cusno`s in subscriptions and in user data.

packages/components/src/Navbar/Component.purs
kaol: I added `toString`. Using plain `show` in code is always suspicious anyway.

ae-mo (approved):

PR #583, opened 2021-03-26
Show temp address changes on their end date but disable removing them

kaol:


apps/mitt-konto/src/MittKonto/Main/Views/UserView/Subscription.purs
```diff
@@ -29,7 +30,7 @@ subscription = make component
       { wrapperProgress: AsyncWrapper.Ready
       , pausedSubscriptions: Nothing
       , pendingAddressChanges: Nothing
-      , now: Nothing
+      , now: unsafeCoerce unit
```
ae-mo, line 33: Is this a good practice? Cause it seems to me that if we do it here there would be no reason why we wouldn't do it in many other places

apps/mitt-konto/src/MittKonto/Main/Views/UserView/Subscription/Helpers.purs
```diff
@@ -55,14 +56,16 @@ translatePaymentMethod paymentMethod =
...
+isPeriodExpired includeCurrentDay baseDateTime endDate =
   case endDate of
     -- If there's no end date, the period is ongoing
     Nothing   -> false
     Just date ->
-      let endDateTime = toDateTime date
-      in maybe true (_ < baseDate) endDateTime
+      let endDateTime = toDate date
+          baseDate = DateTime.date baseDateTime
+          op = if includeCurrentDay then (\a b -> a < b) else (\a b -> a <= b)
```
vapaj, line 67: Is the condition correct? If we want to include current day, shouldn't the check be `a  <= b`?

Also we can ditch the lambdas if it were:
```
op end = if includeCurrentDay then end < baseDate else end <= baseDate
maybe true op endDateTime
```

kaol: I changed the variable name to `excludeCurrentDay` as that was what I meant.

vapaj (approved):

PR #593, opened 2021-04-08
Add test for subscription pause and unpause

kaol:


packages/components/src/PauseSubscription/Component.purs
```diff
@@ -117,13 +117,20 @@ render self =
...
+          , className: "pause-subscription--form"
           , children:
               (case Tuple self.props.oldStart self.props.oldEnd of
                   Tuple (Just start) (Just end) ->
                     [ DOM.text $ "Ursprunglig: " <> formatDateDots start <> " – " <> formatDateDots end ]
                   _ -> []) <>
-              [ startDayInput
-              , endDayInput
+              [ DOM.div
+                  { className: ".pause-subscription--start"
```
vapaj, line 127: I think the name should be without the dot

vapaj (approved):

PR #609, opened 2021-04-21
Use React hooks with Login component

vapaj:


packages/user/src/User/Login.purs
```diff
@@ -20,12 +20,14 @@ import Effect (Effect)
 import Effect.Aff (Aff, error, delay)
 import Effect.Aff as Aff
 import Effect.Class (liftEffect)
+import Effect.Class.Console as Console
```
kaol, line 23: Unused import.

packages/user/src/User/Login.purs
```diff
@@ -34,11 +36,14 @@ import KSF.ValidatableForm as Form
 import React.Basic (JSX)
 import React.Basic.Classic (make)
 import React.Basic.Classic as React
+import React.Basic.Classic as React.Classic
```
kaol, line 39: `React.Basic.Classic` is imported twice as `React` and `React.Classic`. Surely once is enough? And later `React.Basic.Hooks` gets imported as `React`, too.

kaol (changes requested):

kaol (approved):

PR #613, opened 2021-04-22
Premium badge

joakimekroos:


apps/mosaico/src/Mosaico.purs
```diff
@@ -174,28 +174,39 @@ articleList state setState router =
...
-                    ]
+                  [ DOM.div
+                    { className: "mosaico--tag color-hbl"
+                    , children: [ DOM.text $ fromMaybe "" (head a.tags) ]
+                    }
+                  , DOM.h2_ [ DOM.text a.title ]
+                  , DOM.div
+                    { className: "mosaico--article--meta"
+                    , children:
+                      [ case a.premium of
```
vapaj, line 199: We could use [guard](https://pursuit.purescript.org/packages/purescript-prelude/5.0.0/docs/Data.Monoid#v:guard) here:

```purescript
guard a.premium $
  DOM.div
    { className: "mosaico--article--premium background-hbl"
    , children: [ DOM.text "premium" ]
    }
```

It will evaluate a boolean, and if it's true, the given `DOM.div` will be returned. Otherwise it will be `mempty`.

The source for `guard` is basically what you wrote here 😁
```
guard :: forall m. Monoid m => Boolean -> m -> m
guard true a = a
guard false _ = mempty
```

vapaj (approved):

PR #621, opened 2021-04-29
Search with Janrain and Faro results separated

kaol:


packages/components/src/Persona.purs
```diff
@@ -605,8 +586,61 @@ getPayments uuid auth =
...
+type ApiFaroUser =
+  { cusno         :: Cusno
+  , name          :: String
+  , address       :: Nullable Address
+  , email         :: Nullable String
+  , subscriptions :: Nullable (Array Subscription)
+  }
+
+type ApiSearchResult =
+  { janrain     :: Nullable ApiJanrainUser
```
vapaj, line 609: I now wish we had called these `janrainUser` and `faroUsers` 😁 Well, maybe we'll change them at some point.

vapaj (approved):

PR #627, opened 2021-05-03
Add article timestamps

vapaj:


apps/mosaico/src/Article/Article.purs
```diff
@@ -125,6 +134,41 @@ render { props, state, setState } =
...
+        , children:
+            [ publishingTime pubTime
+            , foldMap updateTime maybeUpdateTime
+            ]
+        }
+      where
+        publishingTime time =
+          DOM.div
+            { className: "mosaico--article-published-timestamp"
+            , children: [ DOM.text $ "Pub. " <> formatArticleTime time ]
```
ae-mo, line 164: Is `"Pub."` Swedish?

vapaj: Yep. This follows roughly the same layout as the apps:
https://app-article.ksfmedia.fi/?uuid=356cb203-0d2d-486e-b66d-e59f5e1021bc&paper=hbl&mode=normal

packages/components/src/Lettera/Models.purs
```diff
@@ -18,22 +33,101 @@ isPreviewArticle :: FullArticle -> Boolean
 isPreviewArticle (PreviewArticle _) = true
 isPreviewArticle _ = false
 
-type ArticleStub =
-  { title    :: String
-  , uuid     :: String
-  , preamble :: String
+newtype LocalDateTime = LocalDateTime DateTime
+
+fromUTCTime :: DateTime -> Effect (Maybe LocalDateTime)
```
vapaj, line 38: Do we need a comment here? 🤔

packages/components/src/Lettera/Models.purs
```diff
@@ -18,22 +33,101 @@ isPreviewArticle :: FullArticle -> Boolean
...
+parseArticle = parseArticleWith fromJSArticle
+
+parseArticleStub :: Json -> Effect (Either String ArticleStub)
+parseArticleStub = parseArticleWith fromJSArticleStub
+
+parseDateTime :: String -> Maybe DateTime
+parseDateTime = hush <<< unformat dateTimeFormatter
+
+fromJSArticleStub :: JSArticleStub -> Effect (Maybe ArticleStub)
+fromJSArticleStub jsStub@{ publishingTime } = runMaybeT do
```
ae-mo, line 116: This function and `fromJsArticle` are bound to fail if the timestamps cannot be parsed. I’m not sure this would be the best way to go at it.

What would happen if there was a change in the format of the timestamps? Can we guarantee that it won’t change unexpectedly for the client?

If such a thing happened, no article would be displayed until the problem is fixed.

packages/components/src/Lettera/Models.purs
ae-mo: Yes I think we need to mention about the timezone adaptation explicitly, even though the name is a hint.

packages/components/src/Lettera/Models.purs
vapaj: Good point. Will think about this

vapaj: Decided to turn this into `Maybe LocalDateTime` as defaulting it to some nonsense value on parsing error is pointless.

ae-mo (approved):

PR #628, opened 2021-05-04
Add skapa konto button to search

kaol:


packages/components/src/Search.purs
```diff
@@ -68,9 +79,49 @@ search = do
...
+            Right res -> do
+              let faroCusno x = map _.cusno $ Array.head x.faro
+                  updateWithJanrain :: SearchResult -> SearchResult -> SearchResult
+                  updateWithJanrain x state =
+                    if isNothing state.janrain && faroCusno state == faroCusno x
+                      then x
+                      else state
+              setResults $ (map <<< map) $ updateWithJanrain res
+              setAccountData $ const Nothing
+        createAccountForm = (map <<< map)
```
vapaj, line 117: Would you explain what's going on in this function? 😂 It is beautiful though.

kaol: `accountData`'s type is `Maybe (Tuple Cusno (AsyncWrapper.Progress User.NewCusnoUser)))`, so the `map <<< map` means that it's operating on the `AsyncWrapper.Progress User.NewCusnoUser` part. That matches the fourth argument of `renderEditNewUser`.
`setAccountData` by itself has type `(Maybe (Tuple Cusno (AsyncWrapper.Progress User.NewCusnoUser))) -> Maybe (Tuple Cusno (AsyncWrapper.Progress User.NewCusnoUser)))) -> Effect Unit`. Composing it with `map` peels away those functor layers so that it ends up with just `(User.NewCusnoUser -> User.NewCusnoUser) -> Effect Unit`. That's the third argument of `renderEditNewUser`.

packages/user/src/User/User.purs
```diff
@@ -167,6 +171,38 @@ createUserWithEmail newTemporaryUser = do
           pure $ Left $ UnexpectedError err
     Right user -> finalizeLogin Nothing =<< saveToken user
 
+createCusnoUser :: Persona.NewCusnoUser -> Aff (Either UserError SearchResult)
+createCusnoUser newCusnoUser = do
+  newUser <- try $ Persona.registerCusno newCusnoUser =<< requireToken
+  case newUser of
+    Left err
+      | Just (errData :: Persona.EmailAddressInUseRegistration) <- Api.Error.errorData err -> do
+          Console.error errData.email_address_in_use_registration.description
```
vapaj, line 180: Do we want to send any of these errors to Sentry?

packages/components/src/Search.purs
```diff
@@ -198,9 +249,23 @@ search = do
...
-      foldMap (renderJanrain setActiveUser faro) janrain <>
-      (Array.concatMap (renderFaro loadSubs $ isNothing janrain) faro)
+    renderResult actions { janrain, faro } =
+      foldMap (renderJanrain actions.setActiveUser faro) janrain <>
+      (Array.concatMap (\usr@{ cusno } ->
+                         renderFaro
+                         (actions.loadSubs cusno)
+                         (actions.startCreateAccount usr)
+                         (isNothing janrain)
+                         ((fst <$> actions.createAccountForm) == Just cusno)
```
vapaj, line 259: Maybe a matter of taste, but could we indent these one step to the right?

packages/user/src/User/User.purs
kaol: Maybe? I don't know, any users of this function would be internal users and if they see something wrong they'd know to contact us.

vapaj: Fair enough. Don't know if logging to the console is of any use though 🤔

kaol: It's to us if something goes wrong and we try to debug it.

vapaj: Well, they are useful only for developers and only while developing the app. If they were log messages that actually needed our attention, the correct place to send them would be Sentry.

vapaj (approved):

packages/user/src/User/User.purs
vapaj: And I'm not denying their usefulness for developers. Just that in production they're basically just noise.

PR #629, opened 2021-05-06
Feature/add corona banner to app article

MicciKarkia:


apps/app-article/src/App.js
```diff
@@ -77,11 +89,19 @@ class App extends Component {
         if (getUrlParam().has('login')) {
             this.setState({ forceLoginView: true });
         } else {
-            this.getArticle();
-            this.getMostReadArticles();
+            this.getArticle(),
```
vapaj, line 92: Thought this was a syntax error but apparently it's not 😄 Why this change though?

apps/app-article/src/App.js
```diff
@@ -77,11 +89,19 @@ class App extends Component {
...
+            this.getMostReadArticles() 
         }
     }
-    componentDidUpdate() {
+    componentDidUpdate(prevProps, prevState) {
+        const hasCoronaTag = this.state.tags.some(tag => tag.match(/coronavirus/i))
+        const hadCoronaTag = prevState.tags.some(tag => tag.match(/coronavirus/i))
+        console.log('hasCoronatag: ', hasCoronaTag)
```
vapaj, line 99: Remove `console.logs`

apps/app-article/src/App.js
```diff
@@ -19,6 +19,9 @@ import ManuallyRelatedArticles from "./components/manually-related-articles";
 import Cookies from 'js-cookie';
 import { AndroidView } from 'react-device-detect';
 import { Sentry } from './sentry'
+import CoronaBanner from '../../../packages/components/src/CoronaBanner/banner'
+import { someAuth } from '@ksf-media/user/build';
```
vapaj, line 23: Is this needed?

apps/app-article/src/App.js
MicciKarkia: This should NOT be there!

vapaj:
We should reconsider these rules in `corona-banner.less` as they will lead to (probably) unwanted effects in the application that uses `CoronaBanner` 🙂
```
html,
body {
  padding: 0;
  margin: 0;
  text-align: center;
  font-family: 'Roboto', sans-serif;
  font-size: 16px;
}

a {
  color: inherit;
  text-decoration: none;
}

* {
  box-sizing: border-box;
}
```

apps/app-article/src/App.js
```diff
@@ -157,6 +173,24 @@ class App extends Component {
...
+                        newCases: data.newCases,
+                        hospitalised: data.hospitalised,
+                        deaths: data.deaths,
+                        vaccinated: data.vaccinatedAmount,
+                        vaccinatedPercentage: data.vaccinatedPercentage
+                    }
+                    })
+                })
+                .then(() => {
+                    this.setState({ showBanner: true })
```
vapaj, line 190: Could this be part of the `setState` above?

apps/app-article/src/App.js
MicciKarkia: No, deleted it.

packages/components/src/CoronaBanner/banner.js
```diff
@@ -0,0 +1,139 @@
...
+                  <div className="stat-label">på sjukhus</div>
+                </div>
+              )}
+              {deaths !== null && (
+                <div className="stat smallest-screens-hidden">
+                  <div className="stat-value">{deaths}</div>
+                  <div className="stat-label">dödsfall</div>
+                </div>
+              )}
+              {(vaccinated !== null || vaccinatedPercentage) && (
```
vapaj, line 63: Do we need this condition in the end?

vapaj: Because in line 66 we are rendering it regardless of what value it has, and in `showVaccinatedPercentage` we actually do some comparison on it too.

packages/components/src/CoronaBanner/banner.js
```diff
@@ -0,0 +1,139 @@
...
+        </em>
+      </div>
+    </div>
+  )
+}
+
+export default CoronaBanner
+
+/*
+{vaccinated !== null && vaccinatedPercentage && (
```
vapaj, line 122: I think you forgot this 🙂

packages/components/src/CoronaBanner/banner.js
MicciKarkia: I was trying to make it so that if the vaccinated amount is missing but we have the percentage, then only percentage shows without parentheses around the value. But now I see what you mean, I will clean it up and make a new try!

MicciKarkia: My goal is to cover all 4 possible outcomes:

1. amount === null && percentage === null
(<div className="stat"> element won't render at all)

2. amount !== null && percentage !== null
(<div className="stat"> element will render and inside it <div className="stat-value">  will show amount and inside parentheses percentage number and sign, and <div className="stat-label"> will render)

3. amount !== null && percentage === null
(<div className="stat"> element will render and inside it <div className="stat-value"> will only show amount but not percentage number or sign or parentheses and <div className="stat-label"> will render)

4. amount === null && percentage !== null
(<div className="stat"> element will render and inside it <div className="stat-value"> will only show percentage number and sign without parentheses and <div className="stat-label"> will render)

vapaj: Right, I see. Can we check for truthy values in line 63 or do we need to check explicitly for `not null` (`vaccinated || vaccinatedPercentage`)?

Also, I wish we had pattern matching in javascript 😅

vapaj (approved):

PR #634, opened 2021-05-10
Password reset route

kaol:


packages/components/src/Password/Reset.purs
```diff
@@ -0,0 +1,68 @@
...
+import KSF.Password.SendLink (requestResetLink)
+import KSF.Password.Change (updatePasswordForm)
+import KSF.Tracking as Tracking
+import KSF.User (User)
+import KSF.User as User
+import React.Basic.Hooks (Component, component, useEffectOnce, useState', (/\))
+import React.Basic.Hooks as React
+import React.Basic.DOM as DOM
+
+foreign import getCode :: Effect String
```
vapaj, line 19: Could we use https://pursuit.purescript.org/packages/purescript-web-url/1.0.2/docs/Web.URL.URLSearchParams#v:get ?

packages/components/src/Password/SendLink.purs
```diff
@@ -0,0 +1,109 @@
...
+        renderSuccess =
+          DOM.div
+            { className: "success-text"
+            , children: [ DOM.text "Vi har skickat ett e-post med instruktioner för att skapa ett nytt lösenord." ]
+            }
+
+        renderFailure =
+          DOM.div
+            { className: "error-text"
+            , children: [ DOM.text "Något gick fel." ]
```
vapaj, line 98: Should we have a way to recover from this failure, e.g. "try again"?

packages/components/src/Persona.purs
```diff
@@ -121,6 +121,24 @@ requestPasswordReset :: String -> Aff Unit
 requestPasswordReset email = do
   callApi accountApi "accountPasswordForgotPost" [ unsafeToForeign { email } ] {}
 
+type TokenResponse =
```
vapaj, line 124: This is probably not used after the password reset change we made to Persona 🙂

packages/components/src/Password/Change.purs
```diff
@@ -0,0 +1,165 @@
...
+                , errorView: const renderFailure
+                }
+            ]
+        }
+      where
+        renderEdit :: PasswordReset -> JSX
+        renderEdit { password, confirmPassword } =
+          DOM.form
+            { className: "password--reset"
+            , id: "janrain-password-form"
```
vapaj, line 101: Maybe leave "janrain" out. Would `"password-reset-form"` work?

vapaj (approved):

PR #636, opened 2021-05-14
Show descriptions of each package in subscriptions

kaol:


apps/mitt-konto/src/MittKonto/Main/Views/UserView/Subscription.purs
```diff
@@ -50,7 +50,9 @@ render self@{ props: { now, subscription: sub@{ package, paymentMethod, state }
     (DescriptionList.descriptionList
          { definitions:
              [ { term: "Produkt:"
-               , description: [ DOM.text package.name ]
+               , description: DOM.text package.name Array.:
```
vapaj, line 53: I think `cons` reads a bit better than `:` in this case
```purescript
DOM.text package.name `cons`
...
```

vapaj (approved):

PR #640, opened 2021-05-17
Proposed Prettier formatting rules, and example files with formatting applied

danielrlc:
## Example files with Prettier formatting applied

I've applied Prettier formatting to a number of files using the `prettierrc` configuration file I've added here, so you can see the kind of changes Prettier makes. I'll remove these changes before merging to keep this pull request clean.

The Prettier default settings are good, I think, but I suggest we make at least one change, just so we know our `prettierrc` configuration file is working. My proposed change is enforcing single quotes. But of course we can easily [change several other rules if we like](https://prettier.io/docs/en/options.html).

## Quotes

Prettier handles single and double quotes very skilfully, so we don't need to worry about it.

> If the number of quotes outweighs the other quote, the quote which is less used will be used to format the string - Example: "I'm double quoted" results in "I'm double quoted" and "This \"example\" is single quoted" results in 'This "example" is single quoted'.
[link](https://prettier.io/docs/en/options.html#quotes)

This has been my experience too. I've never had to worry about a Prettier-formatted string breaking from mixed use of ' and ".

## Arrow function parentheses

Prettier's advice:

> At first glance, avoiding parentheses may look like a better choice because of less visual noise. However, when Prettier removes parentheses, it becomes harder to add type annotations, extra arguments or default values as well as making other changes. Consistent use of parentheses provides a better developer experience when editing real codebases, which justifies the default value for the option.
[link](https://prettier.io/docs/en/options.html#arrow-function-parentheses)

## 2 spaces

Note that [2 spaces is the default indentation setting](https://prettier.io/docs/en/options.html#tab-width)

Have I missed anything we discussed? Your thoughts on any of this?


vapaj:
Very nice! About the lambda parentheses:
```
 ...it becomes harder to add type annotations, extra arguments or default values as well as making other changes.
```
So basically the reasoning is that when adding arguments to an anonymous function, the diff is messy?
```
// Single argument
const f = a => console.log(a);

// Add another argument
const f = (a, b) => console.log(a, b);
```
I don't think that adds much noise to the diff. However, I get the consistency point and furthermore, if we have a lambda without arguments, we still need to use parentheses:
```
const f = () => console.log("look ma, no arguments!");
```
So the unary lambda is a special case of some sorts. I'm fine with going either direction with this one.

vapaj:
Should we merge this? Is there anything missing? 🙂

danielrlc:
I don't think so, @vapaj. So I'll merge it once the tests pass.

PR #646, opened 2021-05-21
Add header to mosaico

ae-mo:
As per title.

This is intended to be a basic version, so I aim to make it look like [the screens](https://projects.invisionapp.com/share/EQ10U8THSYAG#/screens) (mobile included) but nothing more.

I would also like to clarify how css classes are to be named.


less/mosaico/header.less
```diff
@@ -0,0 +1,67 @@
...
+            .prenumerera-link {
+                border-radius: 3px;
+                background-color: #00a1ab;
+                color: #fff;
+                text-align: center;
+                padding: 3px 3px 1px 3px;
+            }
+        }
+
+        .logo {
```
vapaj, line 37: I'd still argue the use of scoping is unnecessary here. 

Consider a css class
```css
.mosaico {
  h2 {
    color: red;
  } 
}
```
This is pretty obvious what it does. We know there's a default styling for `h2` defined by the browser, but here we explicitly say "hey, we have this known element `h2` in the global scope, but under `mosaico` it should not follow the normal behaviour, but instead it should be red."

If we aren't going to have e.g. `logo` anywhere else in the document, then nesting it is not needed. However, if we have a generic `logo` somewhere, then having additional rules based on scope makes sense. Here we have a base rule for `logo` but change its color depending on where we are using it.

```css
.logo {
  width: 100px;
  height: 100px;
  color: blue;
}
.mosaico {
  .logo {
    color: red;
  }
}
.vetrina {
  .logo {
    color: green;
  }
}
```
Also if `logo` exists only in a specific scope, it's hard to tell which one by reading the DOM.
```html
<div class="a">
  <div class="b">
    <div class="c">
      <div class="d">
        <div class="logo">
```
```css
.b {
  .logo {
    color: yellow;
  }
}
```
If we required `logo` to be a direct child of a class, meaning classes `a`, `b`, `c` and `d` having their own definitions for it, we'd end up in a nightmare.

ae-mo: Yes this makes sense. 

But I'd like to clarify the BEM syntax above all so that I can make less output it. I'd like to understand if we're following the standard or not. It doesn't seem so, which is fine, but then what are the rules? 

When they are defined, there should be a README somewhere with them written in.

ae-mo: I'd still say that we should follow an official standard as it is probably easier for everyone, unless we have specific reasons not to do so.

danielrlc: @ae-mo @vapaj If we do decide to follow a particular BEM standard, I think [CSS Tricks's BEM 101](https://css-tricks.com/bem-101/) article has just the right amount of BEM tips and recommended syntax for us to follow easily, and it's just a 5-minute read. I know other sources go into more depth on BEM, but I think this article hits that nice 80/20 (benefit/effort) sweet spot. I found it an excellent refresher on BEM myself.

ae-mo:
While waiting for the missing account and menu icons, I think we can merge this so I can move on to the menu overlay.

apps/mosaico/src/Mosaico/Header.purs
```diff
@@ -0,0 +1,60 @@
...
+                ]
+            }
+        , DOM.div
+            { className: block <> "__logo"
+            }
+        , DOM.div
+            { className: block <> "__account"
+            , children: [ DOM.text "NAME"]
+            }
+        , DOM.div
```
vapaj, line 44: Should we use `DOM.nav` here? 🤔 https://developer.mozilla.org/en-US/docs/Web/HTML/Element/nav

joakimekroos (approved):

joakimekroos (approved):

PR #647, opened 2021-05-21
Simulate click to directly view as customer

kaol:
This is disgusting and I wouldn't be doing it this way if this wasn't for internal use.

Just using `setHref` will reload the page and discard any personation. Pushing or replacing state with History API will change the address in the address bar but React will not react to that change. I don't think this could be done with any `Router.Link` on the button either since getting the user to personate is an `Aff` and not an `Effect`.

If you have an better idea about how to do this then I'm open to it.

### DIFF
--- a/apps/mitt-konto/src/MittKonto/Main.purs
+++ b/apps/mitt-konto/src/MittKonto/Main.purs
@@ -45,6 +45,7 @@ type ViewComponents =
   , paymentDetailView :: JSX
   , creditCardUpdateView :: User.User -> JSX
   , passwordResetView :: JSX
+  , needRootRedirect :: Boolean
   }
 
 app :: Component {}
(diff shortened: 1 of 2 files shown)


packages/components/src/Navbar/Component.purs
```diff
@@ -180,9 +180,14 @@ customerService isPersonating activeUser = do
       userLink user =
         Router.link
```
ae-mo, line 181: What about using `Router.redirect` instead? It redirects as soon as rendered IIRC.

kaol: I had partial success with it. I could do a redirect with the personation once but if I after that return to the search page and try to select another customer, then it won't do the redirect and stays on the search page.
I'll look a bit more if I could figure out what it'd need.

kaol: I managed to do it with the router, though it felt like I was tricking it to do it the way I wanted.

ae-mo (approved):

PR #662, opened 2021-06-01
Format everything with prettier!

vapaj:


danielrlc (approved):
Looking forward to using Prettier full time after this!

PR #663, opened 2021-06-01
HBL 365 purchase app

kaol:
I didn't try to make it pixel perfect with the existing site and I tried to bring the scrolling header animation to this version but it glitched when I did so I left it out.

The spinner animation uses the exact same orange as the HBL365 background so I transformed it into gray scale.


kaol:
Looks like this pull request somehow didn't get updated when I pushed changes to the branch. I'll close this and make a new one.

PR #664, opened 2021-06-02
HBL 365 purchase app

kaol:
This is https://github.com/KSF-Media/affresco/pull/663 again, I created a new PR since that somehow lost track of updates to its branch.


apps/hbl365/src/Main.purs
```diff
@@ -0,0 +1,165 @@
...
+    [ DOM.header_
+        [ DOM.div
+            { className: "container"
+            , children:
+                [ DOM.a
+                    { id: "logo-container"
+                    , href: "/"
+                    , children:
+                        [ DOM.img
+                            { src: "https://customer.hbl.fi/images/new/app-icon-1024x1024-pixels.png" }
```
vapaj, line 54: We should probably include this image in the repo directly

kaol: In fact I did but this URL was still left to point to the old location. Will fix.

vapaj (approved):

PR #665, opened 2021-06-03
Support for generating dispatch.yaml

mrskug:


.github/workflows/previews.yml
```diff
@@ -73,6 +73,10 @@ jobs:
         run: |2
             ruby deploy.rb app-article-server
             mv apps/app-article-server build/app-article-server
+      - name: Build Server Derp server
```
vapaj, line 76: I guess you need to re-generate these yamls 😁

mrskug: yep

vapaj (approved):

PR #676, opened 2021-06-09
Mosaico/ads

joakimekroos:


apps/mosaico/src/Ads/Ad.purs
```diff
@@ -0,0 +1,60 @@
...
+  { contentUnit :: String
+  }
+
+type State =
+  { populated :: Boolean
+  }
+
+insertIntoBody :: JSX -> Array JSX -> Array JSX  
+insertIntoBody adBox body =
+  case length body of
```
vapaj, line 36: Given that we have only two cases, let's simplify this to basic `if .. else`:
```purescript
if length body > 4
then fromMaybe ...
else body  `snoc` adBox
```

apps/mosaico/src/Article/Article.purs
```diff
@@ -133,6 +134,9 @@ render { props, state, setState } =
         title = fromMaybe mempty $ map _.title props.articleStub <|> map _.title letteraArticle
         tags = fromMaybe mempty $ map _.tags props.articleStub <|> map _.tags letteraArticle
         mainImage = (_.listImage =<< props.articleStub) <|> (_.mainImage =<< letteraArticle)
+        body = map renderElement state.body
+        bodyArrayLength = length body
```
vapaj, line 138: I think this variable is not used, right? 🙂

apps/mosaico/src/Ads/Ad.js
```diff
@@ -0,0 +1,6 @@
+exports.fetchAd = function (contentUnit) {
+  // console.log("Fetching ad ", contentUnit);
```
vapaj, line 2: Maybe we can remove commented line

vapaj (approved):

PR #677, opened 2021-06-10
Patch/mosaico header menu

ae-mo:
Add menu to mosaico header.


danielrlc (approved):
Nice to read through this and gain more familiarity with PureScript syntax.

Is the plan to extract the grid utility classes like these ones:

```
.grid-col-1 {
    grid-column-start: 1;
}

.grid-col-2 {
    grid-column-start: 2;
}
```
... into their own .less file at some point, so they can be reused site-wide?

vapaj:
> Is the plan to extract the grid utility classes like these ones into their own .less file at some point, so they can be reused site-wide?

Yes, exactly. If you wish, you can do it in this branch already @danielrlc if you're continuing the work 🙂 Or we can merge this and do it separately.

BTW this is very neat! 👏

vapaj (approved):

danielrlc:
> Yes, exactly. If you wish, you can do it in this branch already @danielrlc if you're continuing the work 🙂 Or we can merge this and do it separately.

Could you merge this @vapaj, and then I'll make those changes in a new branch?

PR #689, opened 2021-06-17
Set cusno action for search

kaol:


kaol:
Same as #688 but that one wound up having the wrong branch as basis.

mrskug (approved):

PR #691, opened 2021-06-18
Send password reset emails from search

kaol:


packages/components/src/Search.purs
```diff
@@ -63,16 +65,24 @@ type TaggedSubscription =
...
+
+data PersonaUserEdit
+  = SetCusno (AsyncWrapper.Progress (Maybe Cusno))
+  | ControlPassword (AsyncWrapper.Progress EmailPassword)
+
 search :: Component Props
 search = do
   component "Search" \ { setActiveUser, resetRedirect, now } -> React.do
-    query /\ setQuery <- useState' Nothing
+    query /\ setQuery <- useState' $ Just "kari.pahula+demo5@ksfmedia.fi"
```
vapaj, line 80: Is your email here on purpose? 🙂

vapaj (approved):

packages/components/src/Search.purs
kaol: Oops. I got tired of typing it after nth iteration.

PR #708, opened 2021-07-19
Improve mosaico header menu (desktop version)

ae-mo:
As per title, I'm trying to get this closer to [the current design](https://projects.invisionapp.com/share/EQ10U8THSYAG#/screens/454006009).


yarn.lock
```diff
@@ -1756,11 +1756,6 @@ alphanum-sort@^1.0.0:
   resolved "https://registry.yarnpkg.com/alphanum-sort/-/alphanum-sort-1.0.2.tgz#97a1119649b211ad33691d9f9f486a8ec9fbe0a3"
```
vapaj, line 1756: Do you have `package.json` changes uncommited? Just wonderig how `yarn.lock` got modified without them 😄

ae-mo: :thinking:  maybe it's some rebasing magic, or I think I ran `yarn install` a couple of times

vapaj: Running `yarn install` should not make changes to the lock file without changes in `package.json` 🤔

ae-mo: The changes seem to be due to some cleanup, as I only see stuff being removed. Maybe I have deleted (I think I did once) the `node_modules` folder and so **yarn** had to reinstall all the dependencies, which might have changed since last time

vapaj: If dependencies don't change and the lock file is intact, running `yarn install` should not change anything (only install the packages into `node_modules`). I think this file can be reverted to the version `master` has.

If you want to experiment a bit, I'd be interested in the results of:
1. revert `yarn.lock` changes
2. `yarn install`
3. are there any changes to `yarn.lock`?

danielrlc (approved):
I left one specific comment in the code. I also have a more general question, which is whether it's possible for JSX to be used directly with PureScript, instead of the `DOM.div` approach? I'm mentioning it because for me, the `DOM.div` looks so totally different to the final source code, and JSX is a lot closer.

But... I'm guessing even if it's possible, there are trade-offs involved. :)

yarn.lock
ae-mo: > are there any changes to yarn.lock?

Nope 🤔

less/mosaico/header/menu.less
```diff
@@ -1,196 +1,121 @@
-@import (reference) '../../ksf-colors';
+@import (reference) "../../ksf-colors";
+@import (reference) "../../ksf-grid";
 
 .mosaico-header {
+  &__menu {
+    grid-template-rows: repeat(3, auto);
+    grid-template-columns: 0.1fr 1fr 0.1fr;
+    grid-row-start: 4;
+    grid-column: 1 / span 3;
```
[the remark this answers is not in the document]
ae-mo: I tried it but it compresses the entire menu in a rightmost column created out of nowhere just for it, I have no idea why 😅

ae-mo:
> I also have a more general question, which is whether it's possible for JSX to be used directly with PureScript, instead of the `DOM.div` approach? I'm mentioning it because for me, the `DOM.div` looks so totally different to the final source code, and JSX is a lot closer.

To the best of my knowledge it is not possible 😢

PR #713, opened 2021-07-30
Make http requests asynchronous

vapaj:
This makes things nicer and a bit faster too.

### DIFF
--- a/apps/app-article-server/src/server/index.js
+++ b/apps/app-article-server/src/server/index.js
@@ -42,14 +42,15 @@ async function renderArticle(articleId, res, authHeaders, queryParams, queryStri
       return httpStatus < 400 || httpStatus === 403;
     },
   });
-  const userReq = _.get(authHeaders, "authuser")
-    ? axios.get(process.env.PERSONA_URL + "/users/" + authHeaders.authuser)
-    : null;
   const paper = queryParams.paper || "hbl";
   const darkModeEnabled = queryParams.mode === "dark";
   const mostReadReq = axios.get(process.env.LETTERA_URL + "/mostread?paper=" + paper);
 
-  const requests = [articleReq, mostReadReq, userReq].filter((r) => r !== null);
+  const requests = [articleReq, mostReadReq];
+  // If we have a user id in the headers, let's fetch the user too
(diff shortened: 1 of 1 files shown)


apps/app-article-server/src/server/index.js
```diff
@@ -1,27 +1,30 @@
...
+
+require("dotenv").config();
+
 app.use("/dist", express.static(`${__dirname}/../client`));
 
 app.get("/article/:id", async (req, res) => {
   const articleId = req.params.id;
   if (articleId && UUID.validate(articleId)) {
     const authHeaders = () => {
       if (req.headers.authuser && UUID.validate(req.headers.authuser) && req.headers.authorization) {
```
danielrlc, line 23: @vapaj `authUser` rather than `authuser` maybe, to make the naming consistent with `authHeaders` on the line before?

apps/app-article-server/src/server/index.js
```diff
@@ -32,87 +35,93 @@ app.get("/article/:id", async (req, res) => {
...
+	});
+	const markup = ReactDOM.renderToString(articleJSX);
+	const finalHtml = generateHtml(markup, updatedArticle, user);
+	res.send(finalHtml);
+      })
+    )
+    .catch((errors) => {
+      const markup = ReactDOM.renderToString(<ErrorPage message={"Artikeln kunde inte hämtas!"} />);
+      const html = generateHtml(markup);
+      console.warn("Failed to fetch article!", {
```
danielrlc, line 118: Does this `console.warn` need removing?

danielrlc (approved):
Nice to see the familiar axios being used. I like its syntax. :)

apps/app-article-server/src/server/index.js
vapaj: The `req` object comes from [express](https://expressjs.com/), so we don't have control over that unfortunately.

apps/app-article-server/src/server/index.js
vapaj: For javascript running in the browser, one might argue logging to the console is indeed unnecessary in production.

In this case however, the logging is actually useful, as they are saved to google cloud logs.

PR #724, opened 2021-08-18
Require write scope on magic login to Mitt Konto

kaol:


apps/mitt-konto/src/MittKonto/Main.purs
```diff
@@ -77,13 +79,17 @@ app = do
     route /\ setRoute <- useState' initialRoute
 
     useEffectOnce do
-      let attemptMagicLogin =
-            User.magicLogin Nothing \userResponse ->
-              case userResponse of
-                Right user -> do
```
vapaj, line 83: Couldn't we do the `validScope` check in the old callback function, without changing the return type of `magicLogin`?

kaol: Yes, but the callback is an `Effect` and having `magicLogin` just return the `Aff User` felt more natural than doing a second `Aff.launchAff_` in the callback. I would like the code even more if if removed the callback argument from `magicLogin`altogether and just relied on `Aff` being a monad for chaining actions.

vapaj: Right 🤔

> I would like the code even more if if removed the callback argument from magicLogin altogether

Yep, sounds good. Want to do it in this pr?

vapaj (approved):

PR #734, opened 2021-09-01
Export magicLogin in npm package

vapaj:


packages/user/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@ksf-media/user",
-  "version": "0.3.0",
+  "version": "0.3.2",
```
vapaj, line 3: We have `0.3.1` in npm but apparently forgot to bump the version here :smile:

mrskug (approved):

PR #736, opened 2021-09-02
Mosaico ssr improvement

vapaj:
Adds proper server side rendering of Mosaico.

A separate `MosaicoServer` module is made for doing the ssr stuff, as it's difficult to use `Mosaico` directly. This is because we cannot run browser code (we don't have `window` for instance) in the server. `MosaicoServer` is basically just the bare bones version of Mosaico.

The server, if it got an article from Lettera, will write the article json to the object `window.article` by use of a template. Before we render `Mosaico` in the browser, we try to read that article and give it to `Mosaico` as a prop. This will prevent unnecessary requests made to Lettera. Don't know if this is the best way of handling things, as it seems a bit dirty, although cannot think of other solutions to this either at this point.

I also made the HBL logo clickable (will go to the root route). For convenience.


apps/mosaico/src/Article/Article.purs
```diff
@@ -63,45 +60,53 @@ type State =
...
-      { article = Just a
-      , body = map Record.toSum $ _.body $ fromFullArticle a
-      , mainImage = realArticle.mainImage
-      , title = realArticle.title
-      , tags = realArticle.tags
-      , preamble = realArticle.preamble
-      }
+    fullArticle <- affArticle
+    let article = fromFullArticle fullArticle
+    liftEffect $ do
```
kaol, line 101: Could be done without the `$` (or `do` since it's just one statement).

apps/mosaico/src/Article/Article.purs
```diff
@@ -63,45 +60,53 @@ type State =
...
+          , tags: fold $ _.tags <$> article <|>
+                         _.tags <$> props.articleStub
+          , preamble: fold $ _.preamble <$> article <|>
+                             _.preamble <$> props.articleStub
           }
     state /\ setState <- useState initialState
 
     useEffectOnce do
-      loadArticle setState props.affArticle
+      when (isNothing props.article) $ do
```
kaol, line 80: Could be done without the `$` (or the `do` since it's just one statement).

apps/mosaico/src/Article/Article.purs
```diff
@@ -131,9 +136,9 @@ renderImage img =
 render :: Self -> JSX
 render { props, state, setState } =
     let letteraArticle = map fromFullArticle state.article
-        title = fromMaybe mempty $ map _.title props.articleStub <|> map _.title letteraArticle
-        tags = fromMaybe mempty $ map _.tags props.articleStub <|> map _.tags letteraArticle
-        mainImage = (_.listImage =<< props.articleStub) <|> (_.mainImage =<< letteraArticle)
+        title = fromMaybe mempty $  map _.title letteraArticle <|> map _.title props.articleStub
```
kaol, line 139: Extra space before the map (and it doesn't look like it's for aligning lines).

apps/mosaico/src/Article/Article.purs
```diff
@@ -161,37 +166,42 @@ render { props, state, setState } =
         , DOM.ul
             { className: "mosaico-article__some"
             , children:
-              [ DOM.li_
-                [ DOM.a
-                  { href: "#"
-                  , children: [ DOM.img { src: someIcons.facebook } ]
-                  }
+                [ DOM.li_
```
kaol, line 169: This gets repetitive, could it use a function to generate the `li` elements?

apps/mosaico/src/Mosaico.purs
```diff
@@ -66,54 +70,87 @@ routes :: Match MosaicoPage
...
+  useEffectOnce $ locations (routeListener setState) initialValues.nav
+  useEffect state.route do
+    case state.route of
+      Frontpage -> do
+        if null state.articleList
+        then Aff.launchAff_ do
+          frontpage <- Lettera.getFrontpage HBL
+          liftEffect $ setState \s -> s { articleList = frontpage, article = Nothing }
+        -- Set article to Nothing to prevent flickering of old article
+        else liftEffect $ setState \s -> s { article = Nothing }
```
kaol, line 95: Isn't this in `Effect` already? Seems like no-op unless I'm missing something.

apps/mosaico/src/Mosaico.purs
```diff
@@ -66,54 +70,87 @@ routes :: Match MosaicoPage
...
+      ArticlePage _articleId -> pure unit
+    pure mempty
+
+  pure $ render setState state initialValues.nav
+
+routeListener :: ((State -> State) -> Effect Unit) -> Maybe LocationState -> LocationState -> Effect Unit
+routeListener setState _oldLoc location = do
+  case match routes location.pathname of
+    Right path -> setState \s -> s { route = path }
```
kaol, line 104: `setState _ { route = path }`?

kaol (approved):

apps/mosaico/src/Mosaico.purs
vapaj: Yep, you're right.

PR #737, opened 2021-09-03
Prenumerera app

kaol:


.github/workflows/previews.yml
```diff
@@ -115,7 +115,15 @@ jobs:
       - name: Build HBL365
         run: |2
             ruby deploy.rb hbl365
+<<<<<<< HEAD
```
vapaj, line 118: Some merge conflict stuff

kaol: Fixed, and this file should have been generated by `Makefile` in the first place.

But why have we even committed a file that's always autogenerated? But I'm not going to do anything about that in this branch.

vapaj (approved):

PR #739, opened 2021-09-08
Tell user that digi subscriptions are still active during pause

kaol:


apps/mitt-konto/src/MittKonto/Main/Views/UserView/Subscription/Elements.purs
```diff
@@ -372,7 +372,7 @@ pauseSubscriptionComponent self@{ props: props@{ subscription: sub@{ package } }
     , onSuccess: \pausedSubscription ->
                     self.setState _
                       { pausedSubscriptions = toMaybe pausedSubscription.paused
-                      , wrapperProgress = AsyncWrapper.Success Helpers.successText
+                      , wrapperProgress = AsyncWrapper.Success $ map (_ <> "\nDu kan läsa tidningen digitalt utan extra kostnad under uppehållet.") Helpers.successText
```
ae-mo, line 375: Maybe it's worth to modify the [`AsyncWrapper.Success` data constructor](https://github.com/KSF-Media/affresco/blob/71aa3af051d95f54fb27a38177855ad73c6ffe36/packages/components/src/AsyncWrapper.purs#L22) from

```
Success (Maybe String)
```

to

```
Success (Array String)
```

which accomodates more than one line of text nicely., as well as the case where no text needs to be displayed.

This would remove the need for the trick of using the `\n` character for gluing lines together and then separating them again in the `successWrapper` function.

ae-mo (changes requested):

ae-mo (approved):

PR #745, opened 2021-09-13
Add notes how to develop Mosaico

vapaj:


apps/mosaico/README.md
```diff
@@ -13,6 +13,26 @@ yarn start
...
-For this to work, we need to pre-build the client HTML/JS before running the server and also every time a `purs` file is changed (this is the `yarn build` part).
-So you will be restarting the server and building a lot. There might be smoother ways of doing this.
+We can mentally divide Mosaico into two parts: the server code and the browser code. In production, both parts are required in order to run it as designed. However, as building the server dependent parts and restarting the thing continuously takes time and requires patience, it might be desired to run only the browser part when developing Mosaico. That is, if the development work does not concern the server itself. You can start the development server using Parcel with `yarn start-dev`.
+
+When developing the server side bits, you need to restart the server after any changes with `yarn start` or `spago run` (there might be smoother ways of doing things). Note that if your server side code requires also changes to the browser side of things, you need to build the static files with Parcel before running your server. This is what `yarn build` does. This is a bit clumsy and again, a bit time consuming.
```
ae-mo, line 18: > (there might be smoother ways of doing things)

For now the smoothest way I found is to run at the same time these three commands:

```
$ spago run
$ spago build --watch
$ parcel watch web/index.html -d dist/client --no-cache --public-url /assets
```

There is [a bug in **spago**](https://github.com/purescript/spago/issues/506) that, when fixed, would remove the need for `spago build --watch`, but until then `spago run` does not restart the **node** process upon source code changes.

I think this documentation might be a good place to signal this, as rebuilding manually all the time can be really cumbersome.

ae-mo (changes requested):

PR #747, opened 2021-09-15
add 404 response to the server

MicciKarkia:
only handling the 404, not returning a whole page.
the recommended way to handle a 404 as recommended by the author:
https://github.com/hoodunit/purescript-payload/issues/6

### DIFF
--- a/apps/mosaico/src/Main.purs
+++ b/apps/mosaico/src/Main.purs
@@ -15,10 +15,10 @@ import Effect.Class.Console as Console
 import Effect.Uncurried (EffectFn2, EffectFn3, runEffectFn2, runEffectFn3)
 import KSF.Api (Token(..), UserAuth)
 import Lettera as Lettera
+import Lettera.Models (fromFullArticle, articleToJson, isPreviewArticle)
 import Mosaico.Article as Article
 import MosaicoServer as MosaicoServer
 import Node.Encoding (Encoding(..))
-import Lettera.Models (fromFullArticle, articleToJson, isPreviewArticle)
 import Node.FS.Sync as FS
 import Node.HTTP as HTTP
 import Payload.ContentType as ContentType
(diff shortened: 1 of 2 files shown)


apps/mosaico/src/Main.purs
```diff
@@ -70,18 +71,25 @@ spec ::
                 { response :: TextHtml
                 , guards :: Guards ("credentials" : Nil)
                 }
+          , notFound ::
```
ae-mo, line 74: I don't know how you handled this today with @f-f, but I am afraid that just adding the handler to the spec is not enough to get the page to display.

Indeed I tried building your branch locally and for a non existing article code I still get the article page, simply the code hits the case where the article is not present:
Screenshot 2021-09-15 at 16 53 48


What is missing is some sort of redirect to the `notFound` path.

@vapaj ? What do you think?

 If an article is not found when we are already using the client for routing and rendering, then the client could easily route the user to a not found page, but maybe for the case where server side rendering is supposed to happen could be handled with a 30x answer by the server, redirecting to the not found path. 
 Handling it in the client would result in some issues, because as things are now, the `Mosaico` component would try to fetch again the non existent article from **lettera**

ae-mo: If we use this 'hybrid' approach I have highlighted anyway, we need to make sure we display the same not found page both in the client and the server.

f-f: We plan to handle all of that in the next PRs, with an HTML template, a PureScript component, and so on, so we render the same thing in all the "not found cases".
This change only handles the 404 case when hitting a route that doesn't exists, like `/some/nonexistent/route`.
This change was already hard enough, and it's fairly self-contained, so I think it should be already merged

vapaj (approved):

PR #757, opened 2021-09-27
Mosaico/lists

joakimekroos:
- Pass uuid to `renderArticle` to force rendering of a new article.


apps/mosaico/src/Article/Article.purs
```diff
@@ -77,7 +78,7 @@ articleComponent = do
           }
     state /\ setState <- useState initialState
 
-    useEffectOnce do
+    useEffect props.uuid do
```
vapaj, line 81: I suppose this fixed rendering of clicked articles? It's weird, as `props` should not ever change 😄 I mean, if the `props` change, that means the component is re-mounted with another set of `props`.

apps/mosaico/src/Main.purs
```diff
@@ -142,6 +143,7 @@ notFound { params: { path} } = do
           , onLogin: pure unit
           , user: Nothing
           , article: Just notFoundArticle
+          , uuid: ""
```
vapaj, line 146: If there's a case where a value is not there, we should tend to use `Maybe`. So instead of `String` the type should be `Maybe String`.

apps/mosaico/src/Article/Article.purs
vapaj: Or is this some react shenanigans I'm not aware of?

ae-mo: This is the behaviour we observed: `useEffectOnce`'s callback was effectively being called only once, while `useEffect`'s callback was called when the watched properties of the component changed.

If we look at [`useEffectOnce`'s description from **pursuit**](https://pursuit.purescript.org/packages/purescript-react-basic-hooks/7.0.0/docs/React.Basic.Hooks#v:useEffectOnce) it says:

> Like `useEffect`, but the effect is only performed a single time per component instance.

Indeed the component instance is always the same, because what we do when initializing  the `Mosaico` component is effectfully obtaining the `Article` component and saving it in the state of the `Mosaico` component. Then, when the user wants to read a different article, we pass a different set of properties to that same component instance.

Thus `useEffectOnce` cannot work in our case, but `useEffect` does.

apps/mosaico/src/Main.purs
joakimekroos: Changed it to a `Maybe String`

vapaj:
Looks good! Could you fix compiler warnings also?

vapaj (approved):

PR #758, opened 2021-09-28
Make mosaico menu display in a separate page

### LINKED ISSUE #759: Display mosaico menu in a separate route
As per title.

ae-mo:
Closes #759. Related to #771.


apps/mosaico/src/Mosaico/Header/Menu.purs
```diff
@@ -265,10 +271,27 @@ render { props: { visible } } = DOM.div
...
+          , children: [ DOM.div
+                          { className: sectionHeaderClass
+                          , children:
+                              DOM.div
+                                { className: sectionTitleClass
                                 , children: [ DOM.text title ]
-                                }
-                      ] <> (renderSubsection <$> subsections)
+                                } `cons` 
+                                  if length subsections > 0 then
```
vapaj, line 281: We could use `guard` from `Data.Monoid` here:

```purescript
guard (not $ null subsections)
  DOM.div { className: sectionExpanderClass }
```
This is a matter of personal preference of course, but we could also ditch the `snoc` and construct the array like:
```purescript
-- Instead of
...
, children:
    DOM.div
      { className: sectionTitleClass
      , children: [ DOM.text title ]
      } `cons`
      guard (length subsections > 0)
        [ DOM.div { className: sectionExpanderClass  } ]

-- why not
...
, children
   [ DOM.div
      { className: sectionTitleClass
      , children: [ DOM.text title ]
      }
    , guard (length subsections > 0)
        DOM.div { className: sectionExpanderClass  }
    ]
```

Empty `JSX's` does not affect the DOM, meaning `[ DOM.text "yolo", mempty ]` would render only `[ DOM.text "yolo" ]`.

joakimekroos (approved):

PR #762, opened 2021-09-29
Mosaico/grid

danielrlc:
- Use the same main Mosaico class for server and client-side rendering.
- Only show sidebar on frontpage, not on article page.
- Add sidebar to article's body (further down the page).
- Style article according to Steffen's mockups.

### DIFF
--- a/apps/mosaico/src/Article/Article.purs
+++ b/apps/mosaico/src/Article/Article.purs
@@ -175,16 +175,20 @@ render { props, state, setState } =
                               , foldMap articleTimestamps letteraArticle
                               ]
                           }
-                      , DOM.div
-                          { className: "mosaico--article--body "
-                          , children: case state.article of
-                            (Just (PreviewArticle _previewArticle)) ->
-                              paywallFade
-                              `cons` bodyWithAd
-                              `snoc` vetrina
-                            (Just (FullArticle _fullArticle)) ->
-                              bodyWithAd
-                            _ -> mempty
+                    , DOM.div
+                        { className: "mosaico--article--body "
+                        , children: case state.article of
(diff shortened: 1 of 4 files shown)


apps/mosaico/src/Article/Article.purs
```diff
@@ -141,46 +141,58 @@ render { props, state, setState } =
         mainImage = (_.mainImage =<< letteraArticle) <|> (_.listImage =<< props.articleStub)
```
vapaj, line 141: There seems to be something funny going on with the styles: It renders mobile(?) view of the front page, and any subsequent page render, for instance if you go to an article. However, if you land directly on an article it shows desktop styles https://20210929t131447-dot-mosaico-dot-ksf-staging.uc.r.appspot.com/artikel/d6d08500-ddfe-4566-9f1a-8190345831f7

vapaj (changes requested):

apps/mosaico/src/Article/Article.purs
danielrlc: Did this relate to the Lettera problem, or is it something different?

vapaj: Seems not to, still looks weird (at least in my browser)

danielrlc: Ouch. Looks like we might have merged back in from the remote branch some code we deleted. I think this is something to fix first thing at the next mob programming session.

joakimekroos: We noticed it gets different css classes when rendered ssr compared to client side. I thought we fixed that, but apparently not completely.

PR #765, opened 2021-10-01
Add flag askAccountAlways to Vetrina props for forcing account detail…

kaol:
… form

### DIFF
--- a/apps/hbl365/src/Main.purs
+++ b/apps/hbl365/src/Main.purs
@@ -159,6 +159,7 @@ render product =
         , customNewPurchase: Just NewPurchase.render
         , orderSource: PrenumereraSource -- TODO: find out if there is a more suitable value for this
         , subscriptionExists: anotherAccount
+        , askAccountAlways: false
         }
 
     imgLink href alt src =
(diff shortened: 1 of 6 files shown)


packages/vetrina/src/Vetrina/Vetrina.purs
```diff
@@ -450,6 +438,9 @@ mkPurchase self@{ state: { logger } } validForm affUser =
     when (isUserEntitled self.props.accessEntitlements userEntitlements)
       $ except $ Left SubscriptionExists
 
+    when askAccount $
```
vapaj, line 441: This is a neat trick, however let's add a small comment clarifying it.

apps/hbl365/src/Main.purs
```diff
@@ -159,6 +159,7 @@ render product =
         , customNewPurchase: Just NewPurchase.render
```
vapaj, line 159: Let's ask also for the phone number in `Vetrina.Purchase.AccountForm`. 
Furthermore, do we need to remove some options from the country code dropdown?

kaol: We still don't have phone updates in Persona module. Though I can add it in this branch.

kaol: I don't think there's any issues with having a larger list of countries here since this is used only for making digital only subscriptions.

vapaj:
Actually, let's bump the version of vetrina while we're at it.

PR #766, opened 2021-10-01
Use dummy beforeSend with Sentry to see how it behaves

kaol:


kaol:
Let's not, I figured it out without adding this anywhere.

PR #773, opened 2021-10-07
Add missing article body elements

joakimekroos:
- Add missing article body elements `Footnote`, `Question` and `Quote`
- Add styling for them

### DIFF
--- a/apps/mosaico/src/Article/Article.purs
+++ b/apps/mosaico/src/Article/Article.purs
@@ -2,6 +2,7 @@ module Mosaico.Article where
 
 import Prelude
 
+import Bottega.Models.Order (OrderSource(..))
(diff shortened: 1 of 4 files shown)


danielrlc:
Can we consider using selectors like `.mosaico-article__question` instead of `p.question`, to stick with BEM-like CSS and keep the specificity flatter? Although we need to tidy up/improve our class names in Mosaico, I think we've been getting BEM half right so far by sticking with just single classes as selectors, and not using compound selectors like `p.question`.

joakimekroos:
I started `BEMìfying rhe styles, but ended up opening a pandora's box. Even though I was just focusing on the article I ended up with really messy merge conflicts. I'd like to take the BEM thing as a separate thing.

PR #775, opened 2021-10-08
Put client library hashes in a single package

kaol:
Let's see if using spago for the local dependency package works.


vapaj:
Do we need to move `LocalStorage` and `Cusno` modules under `apis`? Also `Lettera` probably should be there too.

kaol:
> Do we need to move `LocalStorage` and `Cusno` modules under `apis`? Also `Lettera` probably should be there too.

`Persona` and `KSF.Api.Subscription` pulled `Cusno` with them. `Persona` requires `LocalStorage`.

The modules in `apis` are all related to using the generated client libraries and we aren't using one for Lettera, so it wouldn't follow from that. But this branch is quite old, I'd have to check again. It wouldn't likely hurt to move it in either case.

PR #798, opened 2021-11-01
add static page route to backend router

MicciKarkia:
add static page route to backend router.
add error module for showing if something went wrong and page needs to be reloaded.

### DIFF
--- a/apps/mosaico/src/Main.purs
+++ b/apps/mosaico/src/Main.purs
@@ -2,13 +2,10 @@ module Main where
 
 import Prelude
 
-import Affjax (get) as AX
-import Affjax.ResponseFormat (string) as AX
-import Affjax.StatusCode (StatusCode(..))
 import Data.Argonaut.Core (Json)
 import Data.Argonaut.Encode.Class (encodeJson)
 import Data.Array (cons, null)
-import Data.Either (Either(..))
+import Data.Either (Either(..), either)
 import Data.Foldable (foldMap)
 import Data.List (List)
 import Data.List as List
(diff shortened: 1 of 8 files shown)


apps/mosaico/src/Main.purs
```diff
@@ -197,6 +211,43 @@ frontpage { htmlTemplate } _ = do
...
+                        , dangerouslySetInnerHTML: { __html: pageContent } 
+                        }
+              Left jsx -> jsx
+      let mosaicoString =
+            DOM.renderToString
+            $ mosaico
+              { mainContent: StaticPageContent staticPageJsx
+              , mostReadArticles
+              }
+      html <- liftEffect do
```
kaol, line 246: This block seems very linear to me. `html` and `mosaicoTemplate` variables could be omitted with some `>>=` application. But it's fine as is, just a comment about how I'd write it.

apps/mosaico/src/Mosaico/StaticPage.purs
```diff
@@ -0,0 +1,34 @@
...
+  let staticPageUrl = "https://cdn.ksfmedia.fi/mosaico/static/" <> pageName <> ".html"
+  res <- AX.get AX.string staticPageUrl
+  pure $ case res of 
+    Right pageContentResponse ->
+      case pageContentResponse.status of
+        StatusCode 200 -> StaticPageResponse { pageName, pageContent: pageContentResponse.body }
+        StatusCode 404 -> StaticPageNotFound
+        _ -> StaticPageOtherError
+    Left _err ->  StaticPageOtherError
+
```
kaol, line 32: Remove trailing white space.

kaol (approved):

PR #812, opened 2021-11-08
Mosaico/tag list

kaol:


apps/mosaico/src/Main.purs
```diff
@@ -224,6 +232,33 @@ frontpage { htmlTemplate } _ = do
             appendMosaico mosaicoString htmlTemplate >>= appendHead windowVars
   pure $ TextHtml html
 
+tagList :: Env -> { params :: { tag :: String }, guards :: { credentials :: Maybe UserAuth } } -> Aff TextHtml
+tagList { htmlTemplate } { params: { tag } } = do
+  articles <- Lettera.getByTag 0 10 (uriComponentToTag tag) HBL
```
vapaj, line 237: Maybe we could have a bit more by default, maybe 20?

apps/mosaico/src/Main.purs
```diff
@@ -224,6 +232,33 @@ frontpage { htmlTemplate } _ = do
...
+                  { frontpageArticles: articles
+                  , onArticleClick: const $ pure unit
+                  , onTagClick: const $ pure unit
+                  }
+          , mostReadArticles
+          }
+  html <- liftEffect do
+            let windowVars =
+                  "<script>\
+                     \window.tagListArticles=" <> encodeStringifyArticleStubs articles <> ";\
```
vapaj, line 256: In addition to writing this to the template, we should also set it as a prop to Mosaico so it doesn't fetch them again when the browser routes kick in.

So changes are needed to the `Mosaico.purs` module and this javascript file, where we actually start the thing: https://github.com/KSF-Media/affresco/blob/master/apps/mosaico/web/index.js#L14-L20

Although, as I am working with the category pages currently, we will have a lot of different named feeds, which are all basically the same thing, so unifying them will probably be a good idea (in another pull request though, maybe in the one I'm working on). In any case, let's add the `tagListArticles` as a prop for now.

vapaj (changes requested):

apps/mosaico/src/Main.purs
kaol: I bumped it to 20.

apps/mosaico/src/Main.purs
kaol: I added `tagListArticles` to `web/index.js` as well.

apps/mosaico/web/index.js
```diff
@@ -17,6 +17,7 @@ function main() {
 	mostReadArticles={window.mostReadArticles || null}
 	frontpageArticles={window.frontpageArticles || null}
   staticPageContent={window.staticPageContent || null}
+	tagListArticles={window.tagListArticles || null}
```
vapaj, line 20: We also need the purescript implementation. These are the `JSProp` of `Mosaico`, and we need to somehow convert that into `Props`. Then, we probably should put the `props.tagListArticles` into the initial state and render them if they exists.

vapaj (approved):

PR #818, opened 2021-11-09
Strip tags and only use textContent in Mosaico Article preambles

kaol:


kaol:
Let's do this on the back end instead.

PR #823, opened 2021-11-11
Mosaico/faq static page

MicciKarkia:
both html and css stripped down as the drop down accordion is removed. only three question categories so far, more to come later. No styling done yet.


apps/mosaico/README.md
```diff
@@ -60,3 +60,6 @@ export ENTITLED_PASSWORD=
 ```
 spago -x test.dhall test
 ```
+
+## Static Pages
+Since the js script does not work if inserted as inner html, therefore the script is in an external file. The js file has identical name as the html file and fetched at the same time as the html file and run after the DOM tree is built.
```
vapaj, line 65: Could you also add comments how to develop them? Do we need to parcel watch something etc?

vapaj: Also where are they located?

MicciKarkia: Done, changes pushed

apps/mosaico/src/Mosaico/Eval.purs
```diff
@@ -0,0 +1,15 @@
...
+
+import Prelude
+
+import Effect (Effect)
+import Effect.Uncurried (EffectFn1, runEffectFn1)
+
+foreign import evalExternalScriptsImpl :: EffectFn1 (Array String) Unit
+
+evalExternalScripts :: Array ScriptTag -> Effect Unit
+evalExternalScripts scriptTags = runEffectFn1 evalExternalScriptsImpl $ map (\(ScriptTag scriptTag) -> scriptTag) scriptTags
```
kaol, line 11: Like I said for `Mosaico.purs`, I'd prefer adding `<script></script>` here.

Also: I looked at how the javascript function is implemented and it's bonkers. It creates an element, inserts this `<script>`-wrapped thing into it, then grabs its `innerHTML` and then calls `eval` on that. Why isn't it just calling `eval` on the string in the first place without creating some "dummy" element for it? But let's leave it like that for now, I guess this has been waiting for long enough as is.

apps/mosaico/static/fragor-och-svar.js
```diff
@@ -0,0 +1,15 @@
+var questions = document.getElementsByClassName("faq__question");
+var i;
+for (i = 0; i < questions.length; i++) {
```
kaol, line 3: This could have been written as just one event handler on `.static-page__faq`. Events bubble to the containing element's event handler and it could catch any clicks to `.faq_question` elements.

apps/mosaico/static/fragor-och-svar.js
```diff
@@ -0,0 +1,15 @@
...
+  questions[i].addEventListener("click", function () {
+    console.log("button clicked");
```
kaol, line 5: Remove log line.

apps/mosaico/static/fragor-och-svar.html
```diff
@@ -0,0 +1,460 @@
...
+          gör det
+          möjligt för oss att existera). Vi markerar sådana artiklar med markering PREMIUM. Du kan
+          teckna en prenumeration för att få tillgång till Premiummaterialet. Gör din beställning på
+          <a href="https://prenumerera.ksfmedia.fi/">prenumerera.ksfmedia.fi</a>.
+        </p>
+      </div>
+      <div class="faq__qa qa__top">
+        <button class="faq__question">Jag är prenumerant, men kommer ändå inte åt
+          Premiuminnehållet. Vad ska jag göra?</button>
+        <p class="faq__answer faq__answer--open answer">Logga in med din e-postadress som är kopplad till
```
danielrlc, line 20: Not sure if the `answer` class is being styled in the less file? I couldn't find it there.

less/mosaico/static-pages/fragor-och-svar.less
```diff
@@ -0,0 +1,52 @@
...
+    margin: 0 auto;
+    max-width: 100%;
+  }
+
+  .faq__answer {
+    display: none;
+    position: absolute;
+  }
+
+  .faq__q:last-child {
```
danielrlc, line 48: Should this class be `faq__qa`? I couldn't find `faq__q` anywhere else.

MicciKarkia: Yes, it's a typo

danielrlc (approved):
Looks good! And nice BEM work. 👍

apps/mosaico/static/fragor-och-svar.html
MicciKarkia: No it's not atm, it was for testing purpose

apps/mosaico/src/Mosaico/Eval.purs
vapaj: > Why isn't it just calling eval on the string in the first place

If we only had the script as a string, this would work. However we are getting strings of script tags, such as `<script>some js</script>`, so we need a way to export that javascript and evaluate it. AFAIK it needs to be done in a silly fashion like inserting it as inner html somewhere first. Also, we might get script tags without inner html, but with a `src` attribute `<script src="/script.js"></script>` where we again need to make it into an element first, fetch the script and evaluate it.

This is designed to work with whatever Lettera gives us as `externalScripts` (full script elements), but of course here we only have the inner html, so we could have another function that `evals` the strings it gets directly.

ae-mo: Or we could use the `HtmlRenderer`I'm developing

apps/mosaico/src/Mosaico.purs
```diff
@@ -193,7 +194,11 @@ mosaicoComponent initialValues props = React.do
         | otherwise ->
           Aff.launchAff_ do
             staticPage <- fetchStaticPage page
-            liftEffect $ setState _  { staticPage = Just staticPage }
+            liftEffect $ setState _  { staticPage = Just staticPage }  
+            case staticPage of
+              StaticPageResponse r
+                | Just p <- r.pageScript -> liftEffect $ evalExternalScripts [ScriptTag $ "<script>" <> p <> "</script>"]
```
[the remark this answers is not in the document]
vapaj: We can't do that as the function expects full script tags as input (which is what we get from Lettera). But, we do actually have `ExternalScript` (https://github.com/KSF-Media/affresco/blob/master/packages/components/src/Lettera/Models.purs#L119) already. We could use that here instead of `ScriptTag` and create a `mkExternalScript`, or something, that takes in a `String` (the script) and wraps it in `<script>x</script>`.

apps/mosaico/src/Mosaico/Eval.purs
MicciKarkia: Would it be possible to keep the code like this for now and maybe later do these changes?

apps/mosaico/static/fragor-och-svar.js
MicciKarkia: Yes, I shall look into that solution, but maybe for now it could be like this?

kaol (approved):

PR #830, opened 2021-11-18
Use session cookies for user auth in server side Mosaico

kaol:
Note for testing: The cookie uses `Secure`, remove that if you're running Mosaico on localhost.


packages/components/src/Auth.purs
```diff
@@ -43,3 +48,20 @@ requireToken =
   loadToken >>= case _ of
     Nothing -> liftEffect $ throw "Did not find uuid/token in local storage."
     Just loginResponse -> pure loginResponse
+
+setAuthCookies :: Aff Unit
```
vapaj, line 52: As this is only for mosaico, we could call it `setMosaicoAuthCookies` to prevent future confusion 🙂

apps/mosaico/src/Main.purs
```diff
@@ -385,14 +394,37 @@ notFound env _ _ = do
...
+  setCredentials credentials <$> pure (Response.ok EmptyBody)
+
+-- Payload 0.4.0 can't do multiple Set-Cookie headers
+setCredentials :: forall a. Maybe UserAuth -> Response a -> Response a
+setCredentials Nothing response = response
+setCredentials (Just { authToken, userId }) (Response response@{ headers }) = Response
+  response { headers = Headers.set "Set-Cookie" cookies headers }
+  where
+    cookies = "CombinedAuth=" <> (\(Token t) -> t) authToken <> "|" <> UUID.toString userId <>
+              "; Secure; SameSite=Strict; HttpOnly; Path=/"
```
vapaj, line 408: Should we set a TTL also?

kaol: I guess. A session cookie wouldn't survive a browser restart. I'll make it a year.

packages/components/src/Auth.purs
kaol: Function name disambiguated.

vapaj (approved):

Issue #771, opened 2021-10-06
Mosaico menu


ae-mo:
The new site needs a menu. The layout should be responsive and it should be displayed in its own route.

PR #832, opened 2021-11-19
Read menu page categories from Lettera

vapaj:
It seems or some categories, for instance teater, Lettera will not return anything. The queries must be fixed there.

### DIFF
--- a/apps/mosaico/src/Mosaico/Header.purs
+++ b/apps/mosaico/src/Mosaico/Header.purs
@@ -108,7 +108,13 @@ render { state: { menuVisible, menuComponent }, props } =
                  else
                    [ searchButton ])
                 <> [ DOM.div_
-                      [ menuComponent { visible: menuVisible }
+                      [ menuComponent
+                          { visible: menuVisible
+                          , categoryStructure: props.categoryStructure
+                          , onCategoryClick: \categoryLabel _url -> do
+                               props.onCategoryClick categoryLabel
+                               writeCategoryRoute categoryLabel
+                          }
                       , DOM.div
                           { className: iconButtonClass <> " " <> menuButtonClass <>
                               if menuVisible then
@@ -139,11 +145,7 @@ render { state: { menuVisible, menuComponent }, props } =
                    ]
             }
         , DOM.div
-            { className: menuOverlayClass <>
-                           if menuVisible then
-                             " " <> visibleMenuOverlayClass
-                           else
-                             mempty
+            { className: menuOverlayClass <> guard menuVisible (" " <> visibleMenuOverlayClass)
             }
         ]
     }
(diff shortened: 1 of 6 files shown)


apps/mosaico/src/Mosaico/Header/Menu.purs
```diff
@@ -116,103 +109,17 @@ render { props: { visible } } = DOM.div
...
-                       }
-                     ]
-                 }
+    mkSection (n /\ acc) (Category c) =
+      let mkSubsection (Category subc) =
+            { title: subc.label, url: "/" <> show subc.label }
+          section =
+            Section $
+              { title: toUpper $ unwrap c.label
+              , modifier: Just $ "--section" <> show n
```
vapaj, line 118: Not sure if the `--section<n>` is used anywhere at this point for the categories, but maybe better to leave in if required by the `Section` type.

ae-mo: I'm pretty sure it's a remnant of a previous version of the menu. I don't see any modifier mentioned in the `menu.less` file for sections.

Also I think that reading the category structure from the server makes it impossible to have ad hoc modifiers without coupling the server and client implementations, in the sens that the server would become aware of the layout that the client wants to render. All this to say that maybe it makes sense to remove that `modifier` field from the data structure :rofl: 
Unless we come up with some non ad hoc modifier we might want to attach on the fly but I can't come up with such an example now.

vapaj: Let's remove it 😁

ae-mo:
Related to #771

apps/mosaico/src/Mosaico/Header/Menu.purs
```diff
@@ -253,8 +152,8 @@ render { props: { visible } } = DOM.div
         renderMenuLayoutElement (Separator modifier) = renderSeparator modifier
 
         renderSection :: Section -> JSX
-        renderSection { modifier, subsections, title } = DOM.div
-          { className: unwords [ sectionClass, sectionClass <> fromMaybe mempty modifier ]
+        renderSection { subsections, title } = DOM.div
+          { className: unwords [ sectionClass, sectionClass ]
```
ae-mo, line 156: The second `sectionClass` element is redundant

apps/mosaico/src/Mosaico/Header.purs
```diff
@@ -108,7 +108,13 @@ render { state: { menuVisible, menuComponent }, props } =
                  else
                    [ searchButton ])
                 <> [ DOM.div_
-                      [ menuComponent { visible: menuVisible }
+                      [ menuComponent
```
ae-mo, line 111: I think we can get rid of this instance of the menu component here, as this is a leftover from back when the menu was still an overlay.

ae-mo (changes requested):

apps/mosaico/src/Mosaico/Header.purs
vapaj: I see 🤔 I'll do a cleanup

vapaj: I guess all the `menuVisible` stuff as well

ae-mo (approved):

PR #833, opened 2021-11-22
Render related articles blocks

kaol:


apps/mosaico/src/Article/Article.purs
```diff
@@ -385,10 +385,23 @@ render { props, state, setState } =
...
         block = "article-element"
         isDiv = isElem "<div"
         isBlockquote = isElem "<blockquote"
         -- Does the string start with wanted element
         isElem elemName elemString =
           Just 0 == String.indexOf (Pattern elemName) elemString
+        renderRelatedArticle { uuid, title } =
+          DOM.li_
+            [ DOM.a
+                { href: "/artikel/" <> uuid
```
vapaj, line 341: Let's prevent default behavior of making a server side call and just write this to the router

less/mosaico/article.less
```diff
@@ -206,7 +206,31 @@
...
-
+  &__related {
+    display: block;
+    margin: 30px auto;
+    padding: 0 30px 0 15px;
+    ul {
+      list-style: none;
+      margin-top: 5px;
+      li {
+        border-left: 4px solid @hbl-orange; // TODO: deal with other site's colours
```
vapaj, line 217: We should probably add some margin to the `li` elements, as per the layout, so we have that small gap between articles

apps/mosaico/src/Article/Article.purs
kaol: What about right click to open in new tab (or middle-click)? We can't get that without using `a href`, the router never sees that action. I think we should support that use.

less/mosaico/article.less
kaol: I already had `padding-bottom` for them, I changed it to `margin-bottom`. No material difference in this case.

apps/mosaico/src/Article/Article.purs
vapaj: We don't need to remove the `a` tag, we just need to prevent the default behavior when we have js available. It does not seem to affect to middle clicks.
https://pursuit.purescript.org/packages/purescript-react-basic-dom/4.2.0/docs/React.Basic.DOM.Events#v:capture
```
DOM.a
  { href: "/artikel/" <> uuid
  , children: [ DOM.text title ]
  , onClick: capture_ $ Console.log "click"
  }
```

kaol: Ok, right. I'll add `onArticleClick`, similar to what `onTagClick` is.

vapaj (approved):

PR #842, opened 2021-11-24
Smart ad placement in articles

danielrlc:
- No ads in the middle of very short articles
- 1 ad in medium length articles
- 2 ads in long articles
- Ads should place themselves between two `<p>` elements – and not just before or after a heading
- Ads are not styled or properly responsive yet


packages/components/src/Lettera/Models.purs
```diff
@@ -276,6 +277,7 @@ data BodyElement
   | Footnote String
   | Question String
   | Quote String
+  | Ad String
```
vapaj, line 280: Let's add a comment here that this block doesn't come from Lettera

apps/mosaico/src/Article/Article.purs
```diff
@@ -392,3 +396,33 @@ render { props, state, setState } =
...
+          insertAt (findAdSpace bodyWithAd $ 2 * elements/3) ad2 bodyWithAd
+        else if elements > 6
+          then fromMaybe body $ insertAt (findAdSpace body $ elements/2) ad1 body
+          else body `snoc` ad1
+      where
+        elements = length body
+        findAdSpace :: Array BodyElement -> Int -> Int
+        findAdSpace body' i 
+          | i > elements = elements
+          | otherwise    = case body' !! (i-1) of
```
vapaj, line 417: Would be perhaps a bit neater if we wrote everything as guards:
```
  findAdSpace :: Array BodyElement -> Int -> Int
  findAdSpace body' i
    | i > elements = elements
    | Just (Html _) <- body' !! (i-1)
    , Just (Html _) <- body' !! (i) 
    = i
    | otherwise = findAdSpace body' (i+1)
```

You could of course write that new guard in one line, but usually it's more readable when separating the pattern matching to new lines:
```
| Just (Html _) <- body' !! (i-1), Just (Html _) <- body' !! (i) = i
```

packages/components/src/Lettera/Models.purs
danielrlc: Yes, so I pushed this comment. Is it clear enough?

```
  -- Note that Ad does NOT come from Lettera, but was added here to make smart ad placement possible
  | Ad String
```

(So @vapaj did catch this, as you thought he would @ae-mo 😄 )

apps/mosaico/src/Article/Article.purs
```diff
@@ -392,3 +396,33 @@ render { props, state, setState } =
...
+              Just (Html _) -> i
+              _             -> findAdSpace body' (i+1)
+            _             -> findAdSpace body' (i+1)
+
+    removeBodyErrors :: Array (Either String BodyElement) -> Array BodyElement
+    removeBodyErrors body = 
+      let partitioned = partition isRight body
+      in removeEither <$> partitioned.yes
+      where
+        removeEither = fromRight (Html "")
```
vapaj, line 428: By first looking at this I thought it turns every `Left String` into empty `Html` blocks but by reading the entire function, that isn't the case, right? 😄

I'm personally annoyed every time I need to do the same thing, as there are no `takeRights` or `takeLefts` functions easily available (in Data.Either or similar). As this is a recurring problem, we should probably consider using something like [Data.Filterable](https://pursuit.purescript.org/packages/purescript-filterable/4.0.1/docs/Data.Filterable#v:partitionMap). No need to do it in this PR though.

Anyway, the solution is fine as the empty `Html` blocks should never get created, but I think there are better ways of solving this with for instance `foldl`. Want to give it a try as an exercise? 🙂

packages/components/src/Lettera/Models.purs
vapaj: Yes, perfect 😅

apps/mosaico/src/Article/Article.purs
danielrlc: Definitely looks cleaner. I just made the change, tested and pushed. So does the comma work like a logical AND there then?

apps/mosaico/src/Article/Article.purs
danielrlc: Okay sure. Best as part of this PR or a later one?

apps/mosaico/src/Article/Article.purs
vapaj: Yes, kind of! You can read a bit on guards and pattern guards [here](https://github.com/purescript/documentation/blob/master/language/Pattern-Matching.md#guards)

apps/mosaico/src/Article/Article.purs
```diff
@@ -392,3 +396,32 @@ render { props, state, setState } =
...
+          bodyWithAd <- insertAt (findAdSpace body $ elements/3) ad1 body
+          insertAt (findAdSpace bodyWithAd $ 2 * elements/3) ad2 bodyWithAd
+        else if elements > 6
+          then fromMaybe body $ insertAt (findAdSpace body $ elements/2) ad1 body
+          else body `snoc` ad1
+      where
+        elements = length body
+        findAdSpace :: Array BodyElement -> Int -> Int
+        findAdSpace body' i
+          | i > elements = elements
```
vapaj, line 416: Shouldn't this `elements` be based on the length of the `Array BodyElement` given to `findAdSpace`? 🤔

vapaj: I see, there are calculations done based on that elsewhere too.

vapaj: Anyway, it's the length of the "original" body, and for instance on line 408 we are giving `findAdSpace` an array with assumably different count of elements than the original body. 
```
insertAt (findAdSpace bodyWithAd $ 2 * elements/3) ad2 bodyWithAd
```

Probably not a problem.

vapaj (approved):

PR #845, opened 2021-11-24
Add autoComplete tags for login dialog inputs

kaol:


packages/components/src/InputField/InputField.purs
```diff
@@ -95,6 +100,7 @@ render self@{ props, state } =
                 then "input-field--invalid-field"
                 else mempty
             , disabled: props.disabled
+            , autoComplete: if props.autoComplete == "" then unsafeCoerce Nullable.null else props.autoComplete
```
vapaj, line 103: What's the default behavior if `autoComplete` is missing (is it `"off"`)? Should we default to that instead of `null`?

kaol: What happens is that `purescript-react-basic-dom` interprets `null` the same way as if the value was not defined at all. That can't be expressed without using `unsafeCoerce`. Using this results the same DOM as if we never even defined `autoComplete` at all. It was either this or defining `autoComplete` as `Maybe String` for `InputField` and that would've made it more cumbersome.

Leaving out the `autoComplete` attribute gives the element the default behavior. It already does what you ask for.

vapaj: > It already does what you ask for

I know, I'm just not that comfortable unsafeCoercing stuff where it could be avoided.

vapaj (approved):

PR #849, opened 2021-11-26
Create ad slots in app articles for Google Ad Manager (GAM) to target

danielrlc:
- Top ad slot (MOBPARAD)
- Bottom ad slot (MOBNER)
- Ads can be targeted to specific apps using the `paper` URL param: HBL, ON and/or VN
- Feature parity with old `app-article` code


apps/app-article-server/src/browser/ads.js
```diff
@@ -1,3 +1,6 @@
+// This file is currently unused at 2021-11-26, but could be put back into use,
```
vapaj, line 1: Can we remove this file? We have it in git history if we ever want it back (if we remember it once existed 😅)

apps/app-article-server/src/browser/index.html
```diff
@@ -30,38 +30,95 @@
...
+    <script src="https://securepubads.g.doubleclick.net/tag/js/gpt.js"></script>
     <script>
       var ksf = {};
       ksf.settings = {};
       ksf.settings.brand = {};
       ksf.settings.brand.id = "hbl";
     </script>
-    <script async src="https://script.ksfmedia.fi/app/gamAppAds.min.js"></script>
+
+    <script src="https://securepubads.g.doubleclick.net/tag/js/gpt.js"></script>
```
vapaj, line 53: Looks like the same script we already load in line 45

apps/app-article-server/src/browser/index.html
```diff
@@ -30,38 +30,95 @@
...
+        if (urlParams.has("paper")) {
+          return urlParams.get("paper");
+        }
+        return "hbl";
+      };
+
+      const brandValueName = getBrandValueParam()
+
+      if (brandValueName == "on") {
+        appToTarget = "appON";
```
vapaj, line 70: Seems that this variable not defined anywhere (or is it?). We should probably do it.

apps/app-article-server/src/browser/index.html
danielrlc: Thanks. Fixed.

apps/app-article-server/src/browser/ads.js
danielrlc: Yes, removed. If I use the logic in that file at all, it will be soon, so should be able to remember it's in the git history. 😅

vapaj (approved):

PR #850, opened 2021-11-29
Add user prop to Vetrina to handle logins from the outside

kaol:


packages/vetrina/src/Vetrina/Vetrina.purs
```diff
@@ -62,7 +62,7 @@ type JSProps =
   , paper              :: Nullable String
```
vapaj, line 62: I think this would be a good time to see if the vetrina tests still work 😅

vapaj (approved):

PR #855, opened 2021-12-01
Fix app engine builds (again)

vapaj:
Use `.env.local` for local dev environment. However copy that file to `.env.development` for stage gae builds.


deploy.rb
```diff
@@ -75,6 +76,10 @@ def setup_env(app)
     ENV['NODE_ENV'] = 'production'
   else
     ENV['NODE_ENV'] = 'development'
+    # Copy .env.local to .env.development for Parcel
+    if (File.exist?("#{app['path']}/.env.local"))
+      FileUtils.cp("#{app['path']}/.env.local", "#{app['path']}/.env.development")
```
kaol, line 81: Is symlinking an option?

Also, should `.env.local` be in gitignore?

vapaj: > Is symlinking an option?

Probably, although don't see the difference/benefit as this is ran only in CI basically

> Also, should .env.local be in gitignore?

Let's keep it for developers, there are no secrets there

mrskug (approved):

PR #856, opened 2021-12-01
Add Mosaico.Paper for setting Mosaico wide Paper code

kaol:


apps/mosaico/src/Article/Article.purs
```diff
@@ -275,6 +284,68 @@ render props =
                 }
               ]
           }
+        vnPremium =
```
vapaj, line 287: These could actually live somewhere near `Vetrina`

kaol: I added a new module `Vetrina.Products.Premium` to hold these. Ideally we'd get these from the server.

vapaj (approved):

PR #862, opened 2021-12-08
Add login to header and logout to menu

kaol:


apps/mosaico/src/Mosaico/Header.purs
```diff
@@ -78,15 +61,24 @@ render { props } =
...
             }
-        , DOM.div
-            { className: accountClass
-            , children: [ DOM.text "NAME" ]
-            }
+        , maybe
+            (DOM.div
+               { children: [ DOM.text "LOGGA IN" ]
+               , onClick: handler_ props.onLogin
+               , className: accountClass
```
vapaj, line 68: Could we make the cursor be `pointer` over this?

vapaj (approved):

PR #863, opened 2021-12-09
Set Mosaico user sesssion in cookies and fetch it on server side

kaol:


apps/mosaico/src/Main.purs
```diff
@@ -384,78 +388,68 @@ categoryPage env { params: { categoryName } } = do
...
+
+notFound :: Env -> Maybe (Either Unit User) -> Maybe (Array ArticleStub) -> Aff (Response ResponseBody)
+notFound env user maybeMostReadArticles = do
   let articleJSX =
         Article.render
           { paper: mosaicoPaper
           , article: Right notFoundArticle
           , onLogin: pure unit
-          , user: Nothing
+          , user: join $ hush <$> user
```
vapaj, line 414: Can we bind here: `hush =<< user`?

packages/components/src/Persona.js
```diff
@@ -29,3 +29,9 @@ exports.rawJSONParse = function(x) {
         return value;
     });
 };
+
+exports.constructUserFromObject = function(json) {
+    console.log("constructUserFromObject");
```
vapaj, line 34: Let's remove log lines

packages/components/src/Persona.purs
```diff
@@ -45,6 +45,7 @@ foreign import loginApi :: Api
 foreign import usersApi :: Api
 foreign import rawJSONStringify :: Foreign -> String
 foreign import rawJSONParse :: String -> Foreign
+foreign import constructUserFromObject :: Foreign -> User
```
vapaj, line 48: Does this always succeed?

kaol: It works as long as it uses the same `User` both on the server and the client. I wouldn't use this in the app but I'd expect the web site and Mosaico to stay in close sync. Luckily `User` should be pretty static by now and if we made changes to it we'd need to do it as a multi stage process anyway due to Mitt Konto.

If we wanted to have a better `User` constructor then we should have it in `persona-javascript-client` since I wouldn't want to diverge from what it uses in Mosaico. Or replace other users of its `User` construction as well. This is as good as we have right now.

kaol: Or we could replace it with some lite version of `User` that only handles the fields that Mosaico uses. Does it even use anything other than just the first name? It's not like we'd ever want to see subscriptions, entitlements are going to be checked other way anyway.

vapaj (approved):

PR #875, opened 2021-12-15
Mosaico/menu server

joakimekroos:
- Fix server-side routing for /meny


vapaj:
The server seems to fail with error `ReferenceError: window is not defined` when I try to land on `/meny` 🤔

ae-mo:
> The server seems to fail with error `ReferenceError: window is not defined` when I try to land on `/meny` thinking

Yes I was gonna comment about this, the problem might be [here](https://github.com/KSF-Media/affresco/pull/875/files#diff-672b448a3b2c5735b892b9a58dd45dbd6606ebf0088dcecc91f80a731bcdc073R317):

```
router <- liftEffect makeInterface
```

I think the `router` needs the `window` variable, but of course that's not available in `node`.

vapaj:
> I think the router needs the window variable, but of course that's not available in node.

Right, that's probably it. In `MosaicoServer` we need to create an "empty" router exactly to avoid this situation. We probably need to do it here too. 

https://github.com/KSF-Media/affresco/blob/master/apps/mosaico/src/MosaicoServer.purs#L44-L56

joakimekroos:
Passed an empty router the same way as in
> https://github.com/KSF-Media/affresco/blob/master/apps/mosaico/src/MosaicoServer.purs#L44-L56
Seems to work.

vapaj (approved):

kaol (approved):

PR #882, opened 2021-12-17
Mosaico footer

kaol:


less/mosaico/footer.less
```diff
@@ -0,0 +1,126 @@
...
+
+    text-decoration: none;
+  }
+
+  &__logo-image {
+    background-repeat: no-repeat;
+    background-position: center;
+    background-size: 100%;
+    width: 55px;
+    height: width;
```
MicciKarkia, line 94: This doesn't work, "invalid property value"

less/mosaico/footer.less
```diff
@@ -0,0 +1,126 @@
...
+    grid-row-start: 1;
```
MicciKarkia, line 95: This element is not a grid item

apps/mosaico/src/Mosaico/Footer.purs
```diff
@@ -0,0 +1,60 @@
...
+
+footer :: (String -> EventHandler) -> JSX
+footer onStaticPageClick =
+  DOM.footer
+    { className: "mosaico--footer"
+    , children:
+        [ DOM.div
+            { className: "mosaico-footer__links"
+            , children:
+              [ footerLink "Dataskyddsbrevning" "dataskyddsbrevning"
```
MicciKarkia, line 20: This should probably be Dataskyddsbeskrivning

less/mosaico/footer.less
```diff
@@ -0,0 +1,126 @@
...
+    margin-top: 25px;
+    margin-bottom: 15px;
+  }
+
+  &__other-papers {
+    width: 400px;
+    display: flex;
+
+    @media (max-width: @breakpoint-2col) {
+      width: 250px;
```
MicciKarkia, line 63: At least for the mobile view margin bottom or something like that is needed, otherwise the logos don't show entirely

less/mosaico/footer.less
```diff
@@ -0,0 +1,126 @@
...
+
+  &__logo {
+    display: flex;
+    flex-direction: column;
+    align-items: center;
+
+    justify-content: center;
+
+    &--on {
+      flex: auto;
```
MicciKarkia, line 75: As I understand from https://css-tricks.com/snippets/css/a-guide-to-flexbox/ just having auto is not an option.

less/mosaico/footer.less
```diff
@@ -0,0 +1,126 @@
...
+      mask-image: url("../images/logo-hbl.svg");
+      -webkit-mask-image: url("../images/logo-hbl.svg");
+    }
+  }
+
+  &__logo-caption {
+    font: 400 12px "Roboto", sans-serif;
+    color: @grey-mediumlight;
+
+    justify-self: center;
```
MicciKarkia, line 124: This has no effect since it is not a flex item

apps/mosaico/src/Mosaico/Footer.purs
```diff
@@ -0,0 +1,60 @@
...
+            }
+        , DOM.hr { className: "mosaico-footer__separator" }
+        , DOM.div
+            { className: "mosaico-footer__caption"
+            , children: [ DOM.text "ANDRA KSF-TIDNINGAR" ]
+            }
+        , DOM.div
+            { className: "mosaico-footer__other-papers"
+            , children:
+                [ guard (mosaicoPaper /= VN) $ logo VN
```
vapaj, line 34: Let's render all papers here. There's some comments about this in the issue

less/mosaico/_grid.less
```diff
@@ -15,9 +15,11 @@
     "aside"
     "foot";
 
+  //grid-template-rows: auto auto auto auto auto auto auto 500px;
```
MicciKarkia, line 18: Is this comment supposed to be here?

kaol: Nope. I'll remove it.

joakimekroos (approved):

PR #883, opened 2021-12-17
Mosaico/images

joakimekroos:
I made a new module `Mosaico.Article.Image` which takes some props

- **clickable** does nothing yet. The idea is that it can be set to true if the image is supposed to open full-screen in an overlay when clicked.

- **params** is where you can define the height, width and quality. Aptoma's image thing probably works a bit differently from Imengine, so this needs to be changed accordingly.

- **image** the image provided by Lettera


apps/mosaico/src/Article/Article.purs
```diff
@@ -137,9 +113,10 @@ render props =
                       ]
                   }
             ]
-          , DOM.div
-              { className: "mosaico-article__main-image"
-              , children: [ foldMap renderImage mainImage ]
+          , articleMainImage
+              { clickable: true
+              , params: Just "&width=960&height=540&q=90"
+              , image: fold mainImage
```
vapaj, line 119: Didn't know one could `fold` a record without an explicit `Monoid` instance 😃

Anyway, this must create a record of
```
  { url: ""
  , caption: Nothing
  , thumb: ""
  , alignment: Nothing
  , byline: Nothing
  }
```
and I don't know if we want that. This would, for instance, create urls with only the query params:
```
<img src="&width=960&height=540&q=90"></img>
```

I suggest we don't even call `articleMainImage` if `mainImage` is `Nothing`. We can use `foldMap`, instead of `fold`, for this.

f-f: This is indeed the record that gets created: [`fold` works on a `Monoid` instance](https://pursuit.purescript.org/packages/purescript-foldable-traversable/5.0.1/docs/Data.Foldable#v:fold), which `Record` has, and which will happily call the `Monoid` instance on all its fields, see [here](https://github.com/purescript/purescript-prelude/blob/37717776a3693fd5a0a614239f7de1285cba8cdf/src/Data/Monoid.purs#L98) for the implementation

vapaj (approved):

PR #885, opened 2021-12-20
Webview component

kaol:


apps/mosaico/src/Mosaico.purs
```diff
@@ -26,7 +26,7 @@ import KSF.Auth (enableCookieLogin) as Auth
 import KSF.Paper as Paper
```
vapaj, line 26: Mosaico tries to fecth `/v4beta/list/frontpage?paper=ON&category=Borgå torgkamera` when navigating to the camera. I guess we should prevent it.

kaol: I made it fetch feeds only when category type is feed.

vapaj (approved):

PR #886, opened 2021-12-20
Test most read list

kaol:


apps/mosaico/src/Main.purs
```diff
@@ -182,7 +182,7 @@ main = do
           , menu: menu env
           }
         guards = { credentials: getCredentials, category: parseCategory env }
-    Payload.startGuarded (Payload.defaultOpts { port = 8080 }) spec { handlers, guards }
+    Payload.startGuarded (Payload.defaultOpts { port = 8000 }) spec { handlers, guards }
```
vapaj, line 185: I believe we need to run this at 8080 for app engine 🤔

apps/mosaico/src/Mosaico/Frontpage.purs
```diff
@@ -37,9 +39,12 @@ render props =
       DOM.div
         { className: "mosaico--list-article list-article-default"
         , onClick: props.onArticleClick a
-        , _data: Object.fromFoldable [ Tuple "premium" $ if a.premium then "1" else "0"
-                                     , Tuple "uuid" $ a.uuid
-                                     ]
+        , _data: Object.fromFoldable $
+          -- Known bug, exclude from tests
```
vapaj, line 43: What is this about?

apps/mosaico/src/Main.purs
kaol: Oops. It's a pain that that Mosaico and Lettera share the same default port.

apps/mosaico/src/Mosaico/Frontpage.purs
kaol: I've asked about OC's queries for tags containing colons and nobody seems to know about it so I'll rather just not have tests fail on them.

vapaj (approved):

PR #892, opened 2021-12-23
Use paper param on single article fetch

kaol:


packages/components/src/Lettera.purs
```diff
@@ -74,22 +74,22 @@ handleLetteraError :: LetteraError -> Effect Unit
...
-getArticleWithSlug slug = getArticleWithUrl (letteraArticleSlugUrl <> slug)
+getArticleWithSlug slug = getArticleWithUrl (letteraArticleSlugUrl <> slug) Nothing
 
-getArticleWithUrl :: String -> Maybe UserAuth -> Aff (Either String FullArticle)
-getArticleWithUrl url auth = do
+getArticleWithUrl :: String -> Maybe Paper -> Maybe UserAuth -> Aff (Either String FullArticle)
+getArticleWithUrl url paper auth = do
   let request = AX.defaultRequest
-        { url = url
+        { url = url <> foldMap (\p -> "&paper=" <> Paper.toString p) paper
```
vapaj, line 92: Do we know the query string always exists in the url? Meaning that the first param of `?x=y` is in place.

kaol: Yeah, this is wrong. We know that it doesn't have a query string since the only other query parameter available is "textonly" and Mosaico doesn't use it. It should use "?".

vapaj (approved):

PR #894, opened 2021-12-23
edit html text content, fix answer position

MicciKarkia:


less/mosaico/static-pages/fragor-och-svar.less
```diff
@@ -42,7 +42,7 @@
 
   .faq__answer {
     display: none;
-    position: absolute;
+    position: relative;
```
danielrlc, line 45: Can this line maybe be removed, if it doesn't change anything?

danielrlc (approved):

less/mosaico/static-pages/fragor-och-svar.less
MicciKarkia: Yes, it had been changed and therefore was made redundant, but I forgot to remove it. Have fixed it in the next PR.

PR #896, opened 2021-12-23
Load front page HTML version in setFrontpage

kaol:


ae-mo:
While the html is being fetched a loader is displayed but the most read list is there, only to disappear later when the html is rendered. Maybe something to keep in mind for a future improvement.

ae-mo (approved):

PR #901, opened 2021-12-28
integrate andra laser into main frontpage

ae-mo:
As per title.


apps/mosaico/src/Mosaico/Frontpage/Models.purs
```diff
@@ -0,0 +1,46 @@
...
+andraLaserHook { articles, onClickHandler } = HtmlRenderer.replacingHook
+  { shouldProcessNode: (\n ->
+                          let info = do
+                                name      <- HtmlRenderer.getName n
+                                attribs   <- HtmlRenderer.getAttribs n
+                                className <- attribs.class
+                                children  <- HtmlRenderer.getChildren n
+                                textChild <- head children
+                                text      <- HtmlRenderer.getData textChild
+                                pure $ name /\ className /\ text
```
kaol, line 33: I would rather just use a PureScript record for this instead of nested tuples. `pure {name, className, text}` and then in case `Just {name, className, text}`.

apps/mosaico/src/Main.purs
```diff
@@ -292,20 +295,32 @@ frontpage env { guards: { credentials } } = do
...
-  let frontpage' = Frontpage.render
-                     { content: Just articles
-                     , onArticleClick: const mempty
-                     , onTagClick: const mempty
-                     }
-  let mosaicoString =
+  let mosaico = MosaicoServer.app
+      mostReadArticlesProps = { mostReadArticles, onClickHandler: const $ pure unit }
+      frontpage :: ArticleFeed -> JSX
+      frontpage (ArticleList list) =
```
kaol, line 301: Remove trailing white space. I'm not going to flag all of them this time, there are 16 lines with it, this included. `git diff origin/master origin/mosaico/integrate-andra-laser` marks them with red, at least.

apps/mosaico/src/MosaicoServer.purs
```diff
@@ -17,15 +17,11 @@ import Simple.JSON (write)
 
 type Props =
   { mainContent :: MainContent
-  , mostReadArticles :: Array ArticleStub
+  , mostReadArticles :: JSX
```
kaol, line 20: I don't like this. `JSX` is opaque and using it scatters it all over `Main.purs` what's really rendered in there and it opens up possibilities for bugs since `JSX` could be just anything that's not related to a list of articles whatsoever. If it won't be an `Array ArticleStub` then introduce some other new type for this.

`Main.purs` is overloaded enough even as is. Seeding data types in it is lighter than calling renders.

apps/mosaico/src/Mosaico/Frontpage/Models.purs
```diff
@@ -0,0 +1,46 @@
...
+import Data.Array (head)
+import Data.Maybe (Maybe(..))
+import Data.Tuple.Nested ((/\))
+import Effect (Effect)
+import KSF.HtmlRenderer.Models as HtmlRenderer
+import Lettera.Models (ArticleStub)
+import Mosaico.MostReadList as MostReadList
+
+data Hook
+  = AndraLaser (Array ArticleStub) (ArticleStub -> Effect Unit)
```
kaol, line 14: I'd prefer this to be in English. I'll give a pass to what we're doing with categories since they're ever at most used in simple pattern matching but this use is more involved and I'd need to think about it where it is more and I'd prefer that part to stay monolingual.

apps/mosaico/src/Mosaico/Frontpage.purs
```diff
@@ -10,105 +13,109 @@ import Data.Newtype (un)
...
-import Mosaico.Models (ArticleFeed(..))
+import Mosaico.Frontpage.Models (Hook, toHookRep)
 import React.Basic (JSX)
 import React.Basic.DOM as DOM
 import React.Basic.Events (EventHandler)
 
-type Props =
-  { content :: Maybe ArticleFeed
+-- | Represents a frontpage
+class Frontpage h where
```
kaol, line 26: I don't think type class is the right choice for this. I would find it more idiomatic to replace both
```
newtype ListFrontpage = ListFrontpage ListFrontpageProps
newtype PrerenderedFrontpage = PrerenderedFrontpage PrerenderedFrontpageProps
```

with
```
data Frontpage = ListFrontpage ListFrontpageProps | PrerenderedFrontpage PrerenderedFrontpageProps
```

And replace
```
renderListFrontpage :: ListFrontpageProps -> JSX
renderPrerenderedFrontpage :: PrerenderedFrontpageProps -> JSX
```

with
```
render :: Frontpage -> JSX
```

with pattern matching for `ListFrontpage` and `PrerenderedFrontpage` to give the corresponding render implementation.

kaol (changes requested):

apps/mosaico/src/Mosaico/Frontpage.purs
ae-mo: I was thinking the same

apps/mosaico/src/MosaicoServer.purs
ae-mo: Ah yes it's from a previous iteration where I was breaking everything

apps/mosaico/src/Mosaico/Frontpage/Models.purs
ae-mo: And how does one translate it? `ReadAlso` maybe?

kaol (approved):

PR #907, opened 2022-01-03
add mobile layout for kundservice page

MicciKarkia:
-add html & less file for customer service static page
-mobile layout (draft) so far only, tablet/desktop still to be done

### DIFF
--- a/less/mosaico/static-pages/kundservice.less
+++ b/less/mosaico/static-pages/kundservice.less
@@ -1,0 +1,35 @@
+@import (reference) "../../ksf-colors";
+
+.static-page__customer-service {
+  display: flex;
+  flex-direction: column;
+
+  .customer-service__headline {
+    margin: 20px 0;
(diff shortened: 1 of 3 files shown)


less/mosaico/static-pages/kundservice.less
```diff
@@ -0,0 +1,35 @@
+@import (reference) "../../ksf-colors";
+
+.static-page__customer-service {
+  display: flex;
+  flex-direction: column;
+
+  .customer-service__headline {
+    margin: 20px 0;
+    font-size: 2.2rem;
+    line-height: 2.2rem;
```
danielrlc, line 10: This could maybe be `line-height: 1;` to keep it automatically tied to the `font-size`? (Looks like it would then be 2.2rem * 1 according to here: https://developer.mozilla.org/en-US/docs/Web/CSS/line-height)

danielrlc (approved):

PR #911, opened 2022-01-05
Catch clicks in prerendered html and give them to router

kaol:


apps/mosaico/src/Mosaico/Frontpage/Events.purs
```diff
@@ -0,0 +1,29 @@
...
+import Web.DOM.Node as Node
+
+onFrontpageClick :: (String -> Effect Unit) -> EventHandler
+onFrontpageClick setRoute = capture target $ \tgt -> do
+  let findAnchor :: Node -> Effect (Maybe String)
+      findAnchor node = do
+        let continue = maybe (pure Nothing) findAnchor =<< Node.parentNode node
+        case Node.nodeName node of
+          "A" ->
+            case toMaybe $ (unsafeCoerce node).attributes.getNamedItem "href" of
```
vapaj, line 21: If there is a safe (and not much more complicated) way of reading the "href", why not do it that way?
We only seem to need `Web.DOM.Element.fromNode` and `Web.DOM.Element.getAttribute`.

kaol: If there is, I couldn't find it. In https://pursuit.purescript.org/packages/purescript-web-dom/5.0.0/docs/Web.DOM.Node the PureScript types go only so far as to get a `Node` but there's no functions for accessing its attributes.

kaol: On second look, yes, `Web.DOM.Element`. Could as well do it with functions in it.

vapaj (approved):

PR #913, opened 2022-01-07
Mosaico/achor links for static pages

MicciKarkia:


apps/mosaico/src/Mosaico/Routes.purs
```diff
@@ -48,3 +51,7 @@ routes categories = root *> oneOf
             = pure $ Tuple rs category
             | otherwise = invalid $ free $ Fail "Not a category"
       in Match matchRoute
+    removeHash staticPageRoute =
```
kaol, line 54: A simpler solution would be to just not give the hash to the matcher function in the first place. If the only use for it in it is to remove it.

Let's change `Mosaico.purs` to use `locationState.pathname <> locationState.search` instead of `locationState.path` (in two places).

kaol (changes requested):

kaol (approved):

PR #915, opened 2022-01-10
Add "prerendered" category type

kaol:


apps/mosaico/src/Mosaico.purs
```diff
@@ -161,12 +161,16 @@ mosaicoComponent initialValues props = React.do
...
-          CategoryFeed (Just c)
-            | Just cat <- unwrap <$> Map.lookup c state.catMap
-            , cat.type == Feed -> Just <<< ArticleList <$> Lettera.getFrontpage mosaicoPaper (Just $ unwrap cat.label)
+          CategoryFeed c
+            | Just cat <- unwrap <$> Map.lookup c state.catMap ->
+              let label = unwrap c
+                  getArticleList = ArticleList <$> Lettera.getFrontpage mosaicoPaper (Just label)
+              in case cat.type of
+                Prerendered ->
+                  map Html <<< hush <$> Lettera.getFrontpageHtml mosaicoPaper label >>=
```
vapaj, line 170: It's a matter of taste, but I think this would be easier for the reader if the flow would go in one direction, like
```
map Just <<<  maybe getArticleList pure =<< map Html <<< hush <$> Lettera.getFrontpageHtml mosaicoPaper label
```

vapaj (approved):

apps/mosaico/src/Mosaico.purs
kaol: Now that you mention it, it is questionable style to mix composition directions in the same expression. I changed them to go the same way everywhere and changed the bind direction too.

PR #921, opened 2022-01-13
Cache Lettera responses in server side Mosaico

kaol:


apps/mosaico/src/Cache.purs
```diff
@@ -0,0 +1,155 @@
...
+  { value, newStore } <-
+    case HashMap.lookup key store of
+      Just { validUntil, articles } -> do
+        now <- liftEffect nowDateTime
+        if now > validUntil then fetch else pure { value: articles, newStore: store }
+      Nothing -> fetch
+  AVar.put newStore cache
+  pure value
+
+-- Assumes that only main categories may have prerendered HTML
```
vapaj, line 137: Might not be the case.

packages/components/src/Lettera/Header.purs
```diff
@@ -0,0 +1,30 @@
...
+import Data.Int (fromString)
+import Data.Maybe (Maybe(..))
+import Data.String (Pattern(..), split, trim)
+
+type Directive =
+  { name :: String
+  , value :: Maybe Int
+  }
+
+parseCacheControl :: String -> Array Directive
```
vapaj, line 16: What does the cace control strings look like?

kaol: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control
I'm only using cache-control parsing for Lettera responses which we control so I'm not being overly aggressive with the parsing.

vapaj (approved):

PR #925, opened 2022-01-17
Restore HTML feed to vars and move them after rendered HTML

kaol:
Also retain the same Cheerio template for all content appends and only
serialize it to HTML when done.

### DIFF
--- a/apps/mosaico/src/Main.js
+++ b/apps/mosaico/src/Main.js
@@ -1,18 +1,34 @@
 var cheerio = require("cheerio");
 
-// Writes mosaico html inside #app
-exports.appendMosaicoImpl = function (a, HTML_TEMPLATE) {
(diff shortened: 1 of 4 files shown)


apps/mosaico/src/Main.js
```diff
@@ -1,18 +1,34 @@
...
-  $template("head").append(element);
-  return $template.html();
+exports.appendHeadImpl = function (element, $template) {
+  $template.find("head").append(a);
+  return $template;
 };
 
+// Appends content to body
+exports.appendVarsImpl = function (script, $template) {
+  $template.find("#app-vars").append(script);
```
vapaj, line 29: Do we need to add this to `index.html`?

vapaj (approved):

PR #937, opened 2022-01-27
Add analytics

mfahler9:


apps/mosaico/src/Analytics/Analytics.purs
```diff
@@ -0,0 +1,67 @@
...
+import Data.DateTime (DateTime)
+import KSF.User.Cusno (toString, Cusno)
+import KSF.Api.Subscription (Subscription)
+import Data.Foldable (foldMap)
+import Data.Formatter.DateTime (format)
+import KSF.Helpers (dateTimeFormatter)
+import Data.Array (intercalate)
+
+
+type ArticleMetadata =
```
vapaj, line 18: Do we need this intermediate type? Could we convert `Article` directly into `StringArticleMetadata`?

apps/mosaico/src/Mosaico.purs
```diff
@@ -230,6 +236,27 @@ mosaicoComponent initialValues props = React.do
 
   pure $ render setState state initialValues.components initialValues.nav onPaywallEvent
 
+sendArticleAnalytics:: Article -> Maybe User -> Effect Unit
```
vapaj, line 239: I'd move this function to the `Analytics` module, as it kinda only contains analytics logic (= convert types into analytics types and push to data layer)

apps/mosaico/src/Mosaico.purs
```diff
@@ -136,8 +140,10 @@ mosaicoComponent initialValues props = React.do
           Just uuid -> do
             eitherArticle <- Lettera.getArticleAuth uuid mosaicoPaper
             liftEffect case eitherArticle of
+            --analytiikka tähän
```
vapaj, line 143: This can be removed

apps/mosaico/src/Mosaico.purs
```diff
@@ -230,6 +236,27 @@ mosaicoComponent initialValues props = React.do
...
+          , tags: article.tags
+          , userCusno: case user of
+            Just u -> Just u.cusno
+            Nothing   -> Nothing
+          , userSubs: case user of
+            Just u -> Just u.subs
+            Nothing -> Nothing
+          }
+  pushToDataLayer metadata
+  pure unit
```
vapaj, line 258: `pushToDataLayer` already returns `Effect Unit` so `pure unit` is redundant.

apps/mosaico/src/Analytics/Analytics.purs
```diff
@@ -0,0 +1,67 @@
...
+foreign import _pushToDataLayer :: EffectFn1 StringArticleMetadata Unit
+pushToDataLayer :: ArticleMetadata -> Effect Unit
+pushToDataLayer = runEffectFn1 _pushToDataLayer <<< stringifyMetadata
+
+stringifyMetadata :: ArticleMetadata -> StringArticleMetadata
+stringifyMetadata metadata = 
+  { title : metadata.title
+  , publishingTime: foldMap (\x -> format dateTimeFormatter x) metadata.publishingTime
+  , premium: show metadata.premium
+  , category: fromMaybe "" metadata.category
```
vapaj, line 55: In these kind of situations we could use `fold` (https://pursuit.purescript.org/packages/purescript-foldable-traversable/5.0.1/docs/Data.Foldable#v:fold), which, when given a `Maybe String`, returns the underlying `String`, or if we don't have a value, then `mempty` is returned (which evaluates to `""`).

vapaj: This is fine either way though!

vapaj (approved):

PR #943, opened 2022-01-28
Fix Dataskyddsbeskrivning in footer to correct url

MicciKarkia:


apps/mosaico/src/Mosaico/Footer.purs
```diff
@@ -34,6 +34,11 @@ footer onStaticPageClick =
         ]
     }
   where
+    externalLink caption url =
+      DOM.a
```
vapaj, line 38: Do we want to open it in the same browser tab/window as mosaico?

MicciKarkia: No, it's set to target blank now

vapaj (approved):

PR #958, opened 2022-02-14
Endpoint for invalidating Mosaico's caches

kaol:


packages/components/src/Lettera.purs
```diff
@@ -201,17 +201,15 @@ useResponse f (Right response)
...
 
-getFrontpageHtml :: Paper -> String -> Boolean -> Aff (LetteraResponse String)
-getFrontpageHtml paper category reset = do
+getFrontpageHtml :: Paper -> String -> Maybe String -> Aff (LetteraResponse String)
+getFrontpageHtml paper category cacheToken = do
   let request = AX.defaultRequest
         { url = letteraFrontPageHtmlUrl
                 <> "?paper=" <> Paper.toString paper
                 <> "&category=" <> category
+                <> foldMap ("&cacheToken=" <> _) cacheToken
```
vapaj, line 210: Why this and not Cache-Control?

kaol: I think `Cache-Control` would have been more fitting if this was initiated by Mosaico and not by Lettera like it now is. When Mosaico gets to know to refresh the resource, there's no need to tell Lettera to update its data anymore since it would presumably already have the new data.

I think this could be done just without any query parameter at all as well, since there aren't any caches between Lettera and Mosaico. Though it could be useful for testing, still.

vapaj (approved):

PR #971, opened 2022-03-02
Pass client ip to lettera as header

mrskug:


apps/mosaico/src/Main.purs
```diff
@@ -247,19 +262,19 @@ getDraftArticle env { params: {aptomaId}, query } = do
...
+  -> { params :: { uuidOrSlug :: String }, guards :: { credentials :: Maybe UserAuth, clientip :: Maybe String } }
   -> Aff (Response ResponseBody)
-getArticle env r@{ params: { uuidOrSlug }, guards: { credentials } }
+getArticle env r@{ params: { uuidOrSlug }, guards: { credentials, clientip } }
   | Just uuid <- UUID.parseUUID uuidOrSlug = do
       { user, article, mostReadArticles, latestArticles } <- sequential $
         { user: _, article: _, mostReadArticles: _, latestArticles: _ }
         <$> maybe (pure Nothing) (parallel <<< getUser) credentials
-        <*> parallel (Lettera.getArticle uuid mosaicoPaper r.guards.credentials)
+        <*> parallel (Lettera.getArticle uuid mosaicoPaper r.guards.credentials clientip)
```
kaol, line 272: Let's just use `credentials clientip`. Both variables are in scope and it'd be odd style to to access one via `r` and the other directly.

apps/mosaico/src/Main.purs
```diff
@@ -247,19 +262,19 @@ getDraftArticle env { params: {aptomaId}, query } = do
...
         <*> parallel (Cache.getContent <$> Cache.getMostRead env.cache)
         <*> parallel (Cache.getContent <$> Cache.getLatest env.cache)
       renderArticle env user article mostReadArticles latestArticles
   | otherwise = do
-    article <- Lettera.getArticleWithSlug uuidOrSlug r.guards.credentials
+    article <- Lettera.getArticleWithSlug uuidOrSlug r.guards.credentials clientip
```
kaol, line 277: Ditto.

kaol (approved):
Just a couple of style things.

apps/mosaico/src/Main.purs
mrskug: I was wondering about that

PR #972, opened 2022-03-03
No country preselect

joakimekroos:
- Force user to select country when changing address or making temporary address change.


packages/components/src/CountryDropDown.purs
```diff
@@ -11,27 +12,38 @@ import React.Basic.DOM.Events (targetValue)
...
 countryDropDown countries disabled onChange value =
   DOM.div
     { className: "input-field--container"
     , children:
         [ inputLabel { label: "Land", nameFor: "country" }
         , DOM.select
-            { children: map createOption countries
+            { children: cons placeholderOption $ map createOption countries
             , onChange: handler targetValue onChange
-            , value: fromMaybe "FI" value
```
kaol, line 24: I think it's not right to ignore the `value` parameter like this. Instead, `value` should be set to the placeholder value when its corresponding state is initialized in React. The full life cycle of forms fields with React should have functions also for taking updates from the state, not only updating it.

PR #976, opened 2022-03-03
add iframe to html and script for embedding form

MicciKarkia:


apps/mosaico/static/kontakt.html
```diff
@@ -23,7 +23,19 @@ <h1 class="static-page__headline contact__headline">Kontakta oss</h1>
...
+      onload="window.parent.scrollTo(0,0)"
+      allowtransparency="true"
+      allowfullscreen="true"
+      allow="geolocation; microphone; camera"
+      src="https://form.jotform.com/220593033838356"
+      frameborder="0"
+      style="min-width: 100%; height: 539px; border: none"
+      scrolling="no"
+    >
+    </iframe>
```
kaol, line 38: How about using `<iframe ... />`?

apps/mosaico/static/tipsa-oss.js
```diff
@@ -1,8 +1,88 @@
...
+      }
+    });
+    return result;
+  }
+};
+if (window.addEventListener) {
+  window.addEventListener("message", handleIFrameMessage, false);
+} else if (window.attachEvent) {
+  window.attachEvent("onmessage", handleIFrameMessage);
+}
```
kaol, line 88: Lots of identical content in kontakt.js and tipsa-oss.js. Though I guess we can live with it.

kaol (approved):

apps/mosaico/static/tipsa-oss.js
MicciKarkia: I know, we just have to live with it for now until there is time to fix it

PR #994, opened 2022-03-22
Mosaico/ads

joakimekroos:
- Add some more AdManager content units
- Handle ad refreshing: Refresh ads outside the main content area on every route change, since they are not rendered again after the initial page load.
- Set the ad targeting to testing mode for now. This way we can test different ads using actual content units until Mosaico goes live.


apps/mosaico/src/Mosaico.purs
```diff
@@ -606,5 +627,6 @@ render setState state components router onPaywallEvent =
       router.pushState (write {}) $ "/sök?q=" <> query
 
     simpleRoute path = do
+      runEffectFn1 refreshAdsImpl ["mosaico-ad__top-parade", "mosaico-ad__parade"]
```
kaol, line 630: I'm not quite sure about having this call in this place. Would `routeListener` be better place? That's guaranteed to be called whenever user navigates in any way.

But if this is the way it should work then this place is fine.

kaol (approved):

apps/mosaico/src/Mosaico.purs
joakimekroos: I think that's a good idea. I'll look at it.

PR #1002, opened 2022-03-29
Mosaico/article font size

MicciKarkia:


less/mosaico/article.less
```diff
@@ -110,17 +110,6 @@
       line-height: 1.95rem;
```
danielrlc, line 110: Maybe the `rem` could be removed here? (But not your change, I see.)

less/mosaico/article.less
```diff
@@ -332,7 +307,7 @@
     }
   }
   &__subheadline {
-    font: 600 1rem/1.5rem "Roboto", sans-serif;
+    font: 500 1.07rem/1.5rem "Roboto", sans-serif;
```
danielrlc, line 310: Can the `rem` be removed to make this like line 232 above?

danielrlc (approved):

less/mosaico/article.less
MicciKarkia: Made a Trello card for fixing the line-height as there are several places where rem is used

PR #1005, opened 2022-03-30
Change link to open in same tab

MicciKarkia:


apps/mosaico/src/Mosaico/Footer.purs
```diff
@@ -37,7 +37,6 @@ footer onStaticPageClick =
   where
     externalLink caption url =
       DOM.a
-        { target: "_blank"
         , href: url
```
danielrlc, line 40: Should line 40 start with a `{`?

danielrlc (approved):

apps/mosaico/src/Mosaico/Footer.purs
MicciKarkia: Oh yes, it def should

PR #1007, opened 2022-03-30
Show sleep type for subscription pauses

kaol:


packages/components/src/Api/Subscription.purs
```diff
@@ -75,6 +75,7 @@ data SubscriptionPaymentMethod
 type PausedSubscription =
   { startDate :: JSDate
   , endDate   :: Nullable JSDate
+  , sleepType :: String
```
vapaj, line 78: I would add a type for this rather than using `String`

kaol: I swapped it for a newtype.

vapaj: Why not a sum type though? If we have it in Haskell, we have usually mapped those into purescript as well.

kaol: Because then the representation of `PausedSubscription` wouldn't match with the json received from Persona anymore. We've been using `Subscription` with very light parsing so far and adding a sum type to `PausedSubscription` would cascade all over and we'd likely need to stop using `unsafeFromForeign` in `Mosaico.purs` as well. I think this is the best we can do right now.

vapaj: Fair enough. This is becoming messy and the OpenApiClient is horrible. We both tried rewriting it at some point but maybe we should finally finish it. Also we might want to look at tools to automate haskell -> purs type conversions.

vapaj (approved):

PR #1028, opened 2022-04-11
Add author email field to Author and display it

MicciKarkia:


apps/mosaico/src/Article/Article.purs
```diff
@@ -178,16 +178,23 @@ render imageComponent props =
...
+                    map
+                        (\author -> DOM.div
                           { className: "mosaico-article__author"
                           , children: [ guard (article.articleType == Opinion) $
                                         renderOpinionType article.articleTypeDetails
-                                      , DOM.text authorName
+                                      , DOM.span_ [ DOM.text author.byline ]
+                                      , foldMap
+                                        (\authorEmail -> DOM.span
+                                                        {className: "mosaico-article__author-email"
```
kaol, line 189: Add a space before `className`, our style is to have one after `{`.

apps/mosaico/src/Article/Article.purs
```diff
@@ -178,16 +178,23 @@ render imageComponent props =
...
+                                                        , children: [ DOM.text authorEmail ]
+                                                        }
+                                        )
+                                        author.email
                                       ]
                           })
-                        (_.byline <$> article.authors)
-                    , foldMap
+                        article.authors
+                    <> [foldMap
```
kaol, line 197: How about changing this `foldMap [` to just `map`? I don't have much of preference either way but if it was changed one way on line 181 then let's use the same thing here.

kaol (approved):

apps/mosaico/src/Article/Article.purs
MicciKarkia: fixed!

apps/mosaico/src/Article/Article.purs
MicciKarkia: The reason it's foldMap is to prevent the span to exist if there is no email available (as we only show them for paper specific domains)

MicciKarkia: Will leave it like this for now

PR #1031, opened 2022-04-11
Add contact details for Östnyland

jaakkoha-bytecraft:


apps/mosaico/src/Mosaico/Footer.purs
```diff
@@ -1,61 +1,295 @@
...
+        ]
+    , section "Annonser: "
+        [ tel "09 125 35 58"
+        , DOM.text ", "
+        , email "annons@hbl.fi"
+        ]
+    ]
+
+  secondColumn =
+    [ section "Erja Yläjärvi" [ DOM.text "Chefredaktör och ansvarig utgivare" ]
```
vapaj, line 100: Let's add a whitespace between the title and the content, as in narrower screens these joined together.

Maybe the responsibility of doing so could go to the `section` function even.

vapaj (approved):

PR #1034, opened 2022-04-13
Remove authorization from server-side rendering

jaakkoha-bytecraft:
The backend changes break a couple of things, including at least directly navigating to premium articles as a logged in user.


apps/mosaico/src/Main.purs
```diff
@@ -665,49 +619,46 @@ debugList env { params: { uuid }, guards: { credentials } } = do
...
+          , user: Nothing
           }
 
-categoryPage :: Env -> { params :: { categoryName :: String }, guards :: { category :: Category, credentials :: Maybe UserAuth } } -> Aff (Response ResponseBody)
-categoryPage env { guards: { category, credentials } } = do
+categoryPage :: Env -> { params :: { categoryName :: String }, guards :: { category :: Category} } -> Aff (Response ResponseBody)
+categoryPage env { guards: { category} } = do
   let (Category { label }) = category
-  prerendered <- case credentials of
+  prerendered <- case Nothing of
```
kaol, line 628: Just remove the case statement if the value is constant.

kaol (changes requested):

kaol (approved):

jaakkoha-bytecraft:
This seems to currently break embedded content (eg. corona iframes).

PR #1055, opened 2022-04-27
Parse part blocks and render author lines for letters to the editor

kaol:


kaol:
Since the changes were on top of this and this one was pending, I made this branch render review info blocks as well.

apps/mosaico/src/Article/Renderer.purs
```diff
@@ -0,0 +1,36 @@
...
+import React.Basic.DOM as DOM
+
+renderReviewStars :: String -> JSX
+renderReviewStars content = render { content, hooks: Just [ reviewStarsHook ] }
+
+reviewStarsHook :: HtmlRenderer.HookRep
+reviewStarsHook = HtmlRenderer.replacingHook
+  { shouldProcessNode: \n ->
+     HtmlRenderer.getName n == Just "code" &&
+     HtmlRenderer.getAttrib "data-name" n == Just "stars"
```
vapaj, line 20: Are we trying to parse xml we get from drpublish here? I wouldn't like to scatter logic like this for all lettera's clients, like the mobile apps.

apps/mosaico/src/Article/Article.purs
```diff
@@ -412,3 +473,17 @@ mkShareIcon someName =
         , className: "mosaico-article__some--" <> someName
         }
     ]
+
+groupParts :: Array BodyElement -> Array GroupedElements
+groupParts = Array.reverse <<<
```
vapaj, line 478: Would it be simpler if we had some information already in the block, whether it's a part of some group or not? There's a lot responsibility being left for the clients of Lettera.

kaol: My preferred solution for this would still be to just have a block that could contain a list of other blocks. It would most likely blow up `Lettera.Models` but perhaps it'd be worth it after all.

kaol: I doubt duplicating the part info for each row of it would help any implementations, unless they had a use case for using it on each row independently of others. A basic javascript implementation of this could store the partinfo data to a variable if it went through the array with a for loop, it doesn't seem that difficult to me.

vapaj: > My preferred solution for this would still be to just have a block that could contain a list of other blocks.

I know, and maybe it's the best way of doing it in the end. However, my argument against that is still, that the extra container might be redundant for the UI. The layout requirements for insändare are very light: https://www.hbl.fi/artikel/hoppas-hbl-kan-infria-chefredaktorens-lofte/. If the only way it differs from a regular article, are the bylines and an added `"Svar"` for the reply, do clients need to know it's a special container at all (meaning also the `partStart` and `partEnd` blocks)? Could we simply have a new constructor for `Block` for the bylines?

For reviews, I'd bring back the review boxes we have in Lettera v3. Why did we remove that again? I'm probably forgetting something 😁

apps/mosaico/src/Article/Renderer.purs
kaol: There's a PR on Lettera about this now. I removed the stars generation from this end.

apps/mosaico/src/Article/Article.purs
kaol: It's mostly because review content and letters to the editor were represented in the same way in the source data and I just followed the same logic to represent them on this end as well and didn't even think of Box elements.

And why there's partStart/partEnd blocks comes from that that I had already implemented letters to the editor as inline data and then there was a mysterious ticket about them being done with "paragraph styles" and I added them to be able to apply more styling In Mosaico for whatever that would have required and in the end it meant paragraph styles in the editing system and it wasn't anything for Mosaico after all.

I could as well just revert the whole thing with partStart/partEnd now, none of the requirements that seemed to need them never materialized.

vapaj: Ok, makes sense now. Hope the requirements are set for good at this point 😬

Maybe better to start from a clean slate with this one, I agree.

kaol:
Let's not do it like this, I created yet another branch for this: https://github.com/KSF-Media/affresco/pull/1125

PR #1058, opened 2022-04-28
Prevent React from removing every second class in the footer

jaakkoha-bytecraft:


apps/mosaico/src/Mosaico/Footer.purs
```diff
@@ -258,11 +258,17 @@ column children = DOM.div { className: "mosaico-footer__contact-info__column", c
 
 section :: String -> Array JSX -> JSX
 section title children =
-  DOM.p
-    { className: "mosaico-footer__contact-info__section"
-    , children:
-        [ DOM.b { children: [ DOM.text title, DOM.text " " ] }
-        , DOM.div { children: children }
+  DOM.div -- WTF: this div does nothing... except that it prevents some really weird
```
kaol, line 261: The library has underscored versions of functions that only take a `children` parameter. This could use `DOM.div_` instead.

vapaj (approved):

apps/mosaico/src/Mosaico/Footer.purs
```diff
@@ -258,13 +258,18 @@ column children = DOM.div { className: "mosaico-footer__contact-info__column", c
 
 section :: String -> Array JSX -> JSX
 section title children =
-  DOM.p
```
kaol, line 261: Here's a real explanation of what's going on: https://stackoverflow.com/questions/20134929/div-tag-inside-of-p-looking-weird-in-the-dom-tree

It's not valid HTML to have P contain a DIV. It's not a react but a HTML thing.

PR #1064, opened 2022-05-02
Show an advertorial article below every article

jaakkoha-bytecraft:
There's some slight duplication between Frontpage.render/renderListArticle and renderAdvertorialTeaser (especially over CSS), but I think we should leave that kind of cleanup for after-launch.


apps/mosaico/src/Article/Article.purs
```diff
@@ -276,6 +282,30 @@ render imageComponent props =
         , onTagClick: props.onTagClick
         })
 
+    renderAdvertorialTeaser article =
+      let
+        imgSrc =
+          maybe (fallbackImage props.paper)
+            ((_ <> "&function=hardcrop&width=200&height=200&q=90") <<< _.url)
```
vapaj, line 289: We probably shouldn't add these query params, as they work only with the soon-to-be-deprecated imengine. Instead, we could try the `thumb` of the image, which should be big enough for this purpose.

apps/mosaico/src/Article/Article.purs
```diff
@@ -84,6 +87,7 @@ render imageComponent props =
         bodyWithoutAd = map (renderElement (Just props.paper) imageComponent (Just props.onArticleClick)) body
         bodyWithAd = map (renderElement (Just props.paper) imageComponent (Just props.onArticleClick))
           <<< insertAdsIntoBodyText "mosaico-ad__bigbox1" "mosaico-ad__bigbox2" $ body
+        advertorial = fromMaybe (DOM.div_ []) (renderAdvertorialTeaser <$> props.advertorial)
```
vapaj, line 90: As `JSX` is a `Monoid`, we could use `foldMap` here too:
```
advertorial = foldMap renderAdvertorialTeaser props.advertorial
```

jaakkoha-bytecraft: Thanks, fixed.

vapaj (approved):

PR #1069, opened 2022-05-03
Combine index.js and ads.js to prevent weird race condition.

joakimekroos:
The content of ads.js has been moved into index.js because esbuild was compiling ads.js into a self-executing anonymous function which behaved inconsistently and different across browsers.


danielrlc (approved):
Tested in Firefox, Chrome and Safari. All good.

PR #1074, opened 2022-05-04
Add ads.txt route

vapaj:


apps/mosaico/src/Main.purs
```diff
@@ -350,6 +354,9 @@ frontpageUpdated env { params: { category } } = do
 assets :: { params :: { path :: List String } } -> Aff (Either Failure File)
 assets { params: { path } } = Handlers.directory "dist/assets" path
 
+adsTxt :: forall r. { | r} -> Aff File
+adsTxt = Handlers.file "dist/assets/ads.txt"
```
vapaj, line 358: Don't know if we need it under `dist`. Do we want to serve it from `hbl.fi/assets/ads.txt` too? 🤔

vapaj: Oh, nevermind, I see what you did there! Does this always read our file system on request? I guess it'll be cached in nginx anyway so maybe not a big deal.

joakimekroos: It does always read the file system, yes (unless nginx caches it). Google tries to update the file once every 24 hours according to their docs, so I imagine it won't create a huge amount of extra load.

danielrlc:
> Just a thought I came up with just now: Would Google allow ads.txt to serve redirects instead? We could send their crawler to just grab the file from some bucket instead. Edit: https://stackoverflow.com/questions/47441075/can-you-set-ads-txt-to-redirect-to-a-file-on-another-server says that single hop to an external site is allowed.

Any thoughts on this? My vote is to stick with the solution we have now, as it's working and keeps the ads.txt file closer to hand.

joakimekroos (approved):

PR #1076, opened 2022-05-04
Implement share buttons

jaakkoha-bytecraft:


apps/mosaico/src/Article/Article.purs
```diff
@@ -64,6 +65,7 @@ type Props =
   , user :: Maybe User
   , mostReadArticles :: Array ArticleStub
   , latestArticles :: Array ArticleStub
+  , currentUrl :: String
```
vapaj, line 68: We actually do have the `shareUrl :: Maybe String` in Lettera's `Article` and `ArticleStub` types, however I noticed now they're not exposed in the purescript types. Could we use that instead?

Speaking of which, I'm not sure why it's a `Maybe` in Lettera, as we always know it. I need to look into it.

vapaj: Also manually mapping types from haskell to purescript is a pain 😅

apps/mosaico/src/Share/Share.purs
```diff
@@ -0,0 +1,74 @@
...
+            )
+        , className: "mosaico-article__some--" <> someName
+        }
+    ]
+
+articleShareButtons :: String -> String -> JSX
+articleShareButtons title currentUrl =
+  DOM.ul
+    { className: "mosaico-article__some"
+    , children: if currentUrl /= "" then shareIcons else mempty
```
vapaj, line 61: I would make `currentUrl` a `Maybe String` if it's expected to be sometimes missing.

vapaj:
The share icons of advertorials are messed up for some reason, see for example article 63856950-4f18-470f-bd8e-cec52cfa7e51

apps/mosaico/src/Article/Article.purs
jaakkoha-bytecraft: At least for eg. notFoundArticle in affresco, the shareUrl is quite naturally Nothing. Probably something similar in Lettera's end.

I'll update the code, it should end up being a lot nicer with that change.

jaakkoha-bytecraft:
Oh, one change had actually broken the search page, and it wasn't just tests being flaky.

PR #1080, opened 2022-05-05
Show timestamps for articles in search and category views

jaakkoha-bytecraft:


apps/mosaico/src/Mosaico/Frontpage.purs
```diff
@@ -92,6 +92,10 @@ render (List props) =
                                       { href: "/artikel/" <> a.uuid
                                       , children: [ DOM.h2_ [ DOM.text $ fromMaybe a.title a.listTitle] ]
                                       }
+                                  , DOM.span
+                                    { className: "list-article-timestamp"
```
vapaj, line 96: A styling issue: let's indent two spaces here.

apps/mosaico/src/Mosaico/Frontpage.purs
```diff
@@ -92,6 +92,10 @@ render (List props) =
...
+                                    , children: [ DOM.text $ foldMap formatLocalDateTimeAsArticleTime a.publishingTime ]
```
vapaj, line 97: Nothing wrong with this, however if the only purpose of `formatLocalDateTimeAsArticleTime` is to format the underlying `DateTime` (and we only use it here), we could more simply unwrap it and format in place. As `LocalDateTime DateTime` is a [Newtype](https://pursuit.purescript.org/packages/purescript-newtype/5.0.0/docs/Data.Newtype), it's convenient also:

```
foldMap (formatArticleTime <<< unwrap) a.publishingTime
```

Note that this `DOM.span { className: "list-article-timestamp" ...` is always rendered even if we have no publishing time. This shouldn't ever happen though, as we aren't returning drafts in search results.

kaol:
This uses regular text color for the time stamps, I think `@grey-medium` is the one that should be used instead. And I'm not sure about the font size either, perhaps it should be smaller.

vapaj:
Couldn't find a layout for this, so I propose we merge it and adjust if there's complaints.

vapaj (approved):

PR #1089, opened 2022-05-06
Mosaico/search grid height

kaol:


less/mosaico/search.less
```diff
@@ -82,3 +82,7 @@
 
   margin-bottom: 1em;
 }
+
+.mosaico-search + .mosaico--article-list {
+  margin-top: 30px;
```
MicciKarkia, line 87: For desktop view, this could be like 50px? Because on smaller screens the space between the search bar and article list is much bigger.

MicciKarkia (approved):

PR #1125, opened 2022-05-13
Use box block type parameter and add styles for review box

kaol:
Update Mosaico.Article.Box from React Classic to Hooks style and
remove paper property from it and instead use the ancestor's paper
specific id to select brand colors.

Make Advertorial Standard and Basic components into proper components
and move Image and Box components from Mosaico to them.

### DIFF
--- a/packages/components/src/Lettera/Models.purs
+++ b/packages/components/src/Lettera/Models.purs
@@ -413,6 +413,7 @@ type BoxInfo =
   { title :: Maybe String
   , headline :: Maybe String
   , content :: Array String
+  , type :: String
   }
 
 type QuoteInfo =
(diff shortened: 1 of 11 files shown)


packages/components/src/Lettera/Models.purs
```diff
@@ -413,6 +413,7 @@ type BoxInfo =
   { title :: Maybe String
   , headline :: Maybe String
   , content :: Array String
+  , type :: Maybe String
```
vapaj, line 416: We should always `type` present. According to lettera `title` too actually.

vapaj: (Although didn't check your changes in lettera regarding this yet)

kaol: Yes, I was going to remove the Maybe when Lettera gets updated. The current version of Lettera doesn't have the type field, it would fail all articles with any box blocks if I did it already.

kaol: Lettera has been updated to have the type field, I updated this to remove the Maybe.

vapaj (approved):

PR #1126, opened 2022-05-16
De-state Mosaico variable in Mosaico

kaol:
Remove ads from profile page

### DIFF
--- a/apps/mosaico/src/Mosaico.purs
+++ b/apps/mosaico/src/Mosaico.purs
@@ -93,7 +93,6 @@ type State =
   , categoryStructure :: Array Category
   , catMap :: Categories
   , frontpageFeeds :: HashMap ArticleFeedType FeedSnapshot
-  , showAds :: Boolean
   , ssrPreview :: Boolean
   , advertorials :: Maybe (Array ArticleStub)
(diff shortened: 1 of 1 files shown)


apps/mosaico/src/Mosaico.purs
```diff
@@ -279,39 +271,30 @@ mosaicoComponent initialValues props = React.do
...
             staticPage <- fetchStaticPage page
-            liftEffect $ setState _  { staticPage = Just staticPage, showAds = false }
+            liftEffect $ setState _  { staticPage = Just staticPage }
             case staticPage of
               StaticPageResponse r
                 | Just p <- r.pageScript -> liftEffect $ evalExternalScripts [ScriptTag $ "<script>" <> p <> "</script>"]
               _ -> mempty
-      Routes.DebugPage _ -> pure unit
       Routes.DeployPreview -> liftEffect $ setState _  { route = Routes.Frontpage }
+      _ -> pure unit
```
jaakkoha-bytecraft, line 297: I kinda prefer having explicit match lists than fallbacks in match blocks like this -- it makes it more explicit that we are not doing anything special for this route on purpose.

jaakkoha-bytecraft (approved):

PR #1135, opened 2022-05-18
Refactor and combine server side category render functions

kaol:


apps/mosaico/src/Main.purs
```diff
@@ -857,6 +830,26 @@ htmlContent :: forall a. Response a -> Response a
...
   Response $ response { headers = Headers.set "content-type" ContentType.html response.headers }
 
+stdVars :: Env -> Stamped (Array ArticleStub) -> Stamped (Array ArticleStub) -> Array (Tuple String String)
+stdVars env mostReadArticles latestArticles =
+  [ "mostReadArticles"  /\ encodeStringifyArticleStubs (Cache.getContent mostReadArticles)
+  , "latestArticles"    /\ encodeStringifyArticleStubs (Cache.getContent latestArticles)
+  , "categoryStructure" /\ (JSON.stringify $ encodeJson env.categoryStructure)
+  ]
+
+type WithLists a =
```
jaakkoha-bytecraft, line 840: I wonder if there would be a better name for this, maybe something more verbose like `WithMostReadAndLatestArticles`? Then `paralletWithLists` would probably be something like `parallelFetchArticleMostReadAndLatestArticles` -- which sounds verbose but at least it sounds more specific. The `parallelWithLists` sounds something like a general function doing something similar to `Promise.all`.

kaol: I would never remember a function name like that. I could settle for `parallelWithMosaicoLists`.

PR #1155, opened 2022-05-23
add e-mail address to Tipsa oss page for hbl

MicciKarkia:


apps/mosaico/static/hbl/tipsa-oss.html
```diff
@@ -3,7 +3,7 @@ <h1 class="static-page__headline tip__headline">Tipsa oss</h1>
   <p>Har du ett tips om något vi borde skriva om? Var du på plats när något intressant hände?</p>
   <p>
     Fyll i blanketten nedan och låt oss ta del av det! Lämna gärna dina kontaktuppgifter, så att vi kan kontakta dej vid
-    behov.
+    behov. Du kan även skicka e-post till tipsaoss@hbl.fi.
```
kaol, line 6: Should this use <a href="mailto:">?

jaakkoha-bytecraft: Would be nice to use a link there with `href="mailto:tipsaoss@hbl.fi"`

MicciKarkia: Decided earlier not to use it for a number of reasons, but I can change it.

PR #1161, opened 2022-05-25
Update build systems to esbuild

jaakkoha-bytecraft:
This does the following things:
- Convert all build scripts to use esbuild with a shared config file
- Update docker runtimes from node12 to node16 (because esbuild-plugin-html doesn't support node12)
- Remove all references to parcel


jaakkoha-bytecraft:
Waiting for KSF-Media/docker-diskho/pull/18

PR #1171, opened 2022-05-30
Receive front page updates via Pub/Sub to server side Mosaico

kaol:


apps/mosaico/src/Cache.purs
```diff
@@ -141,26 +161,32 @@ initCache paper categoryStructure = do
   let prerenderedCategories = filter (\(Category c) -> c.type == Prerendered) categoryStructure
       mainFeedCategories = filter (\(Category c) -> c.type == Feed) categoryStructure
   mainCategoryFeed <-
+    (map <<< map <<< dimap identity) (join <<< fromFoldable)
```
vapaj, line 164: `rmap` could replace `dimap identity` here

apps/mosaico/src/Cache/Pubsub.purs
```diff
@@ -0,0 +1,30 @@
...
+
+import Control.Promise (Promise, toAffE)
+import Effect (Effect)
+import Effect.Aff (Aff)
+import Data.Maybe (Maybe)
+import Data.Nullable (Nullable, toMaybe)
+import Lettera.Models (CategoryLabel(..))
+import Lettera (LetteraResponse(..))
+
+foreign import data Message :: Type
```
vapaj, line 13: Was thinking do we need FFI for reading a `Message`, if we just typed it somehow, but what's the type of `message.data`? Google doesn't tell much about it in their docs either 😅 Only `message.data.toString()`.

apps/mosaico/src/Cache/Pubsub.purs
```diff
@@ -0,0 +1,30 @@
...
+
+foreign import subscribeImpl :: (Message -> Effect Unit) -> Effect (Promise Unit)
```
vapaj, line 15: Is it a deliberate decision not to use `Effect.Uncurried` and `Data.Function.Uncurried` with these?

apps/mosaico/src/Cache/Pubsub.purs
kaol: It's a https://nodejs.org/api/buffer.html which I figured out by adding trace statements to everything.

vapaj: Oh, wonderful. Is it actually then https://pursuit.purescript.org/packages/purescript-node-buffer/8.0.0/docs/Node.Buffer ?

apps/mosaico/src/Cache/Pubsub.purs
kaol: All the foreign functions in this module use one parameter so it wouldn't really add anything. I don't know why the module even has `runEffectFn1` since it's pretty much a no op.

apps/mosaico/src/Cache/Pubsub.purs
kaol: I guess, though we'd only use it to call `toString`.

apps/mosaico/src/Cache/Pubsub.js
```diff
@@ -0,0 +1,66 @@
+// TODO is this the way?
```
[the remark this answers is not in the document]
kaol: The reason why there's so much JS code is to get `projectId` defined before deciding whether to run `require('@google-cloud/pubsub')` or use `null`. I'm not quite sure how to do the same thing if these should be PureScript functions instead. Calling PS code from JS is an option, I've read about it in the docs but not tried it for anything yet. Or I could ditch the module-wide `PubSub` variable and pass it along as a function parameter.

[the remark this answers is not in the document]
kaol: I moved most of the credentials manipulation logic inside `Mosaico.Cache.Pubsub.Init`, what's left on the js side is an import to it, one call to a function from it and some constants.

kaol:
Branch moved to private Gitlab.

PR #1183, opened 2022-06-02
Use History back action if prev route exists on menu close

kaol:


jaakkoha-bytecraft (approved):
We should probably consider moving into a modal-style (or a dropdown-style) menu page, this feels like a workaround for functionality around that.

PR #1187, opened 2022-06-03
Show only 5 items in aside lists on mobile

kaol:


less/mosaico/article-list.less
```diff
@@ -231,3 +233,9 @@
     padding: 30px;
   }
 }
+
+.list-article--hide-mobile {
```
jaakkoha-bytecraft, line 237: A simpler way to do this would be something like (within a media query)
```
.mosaico-asidelist.mosaico-asidelist__latest > ul > li:nth-child(5) ~ * {
  display: none;
}
```

jaakkoha-bytecraft (approved):

PR #1190, opened 2022-06-03
Use server loaded article as clickedArticle in initial Mosaico state

kaol:
Collect window variables as Json values instead of encoding them as
Strings first

### DIFF
--- a/packages/components/src/Lettera/Models.purs
+++ b/packages/components/src/Lettera/Models.purs
@@ -225,12 +225,21 @@ type Author =
   , email  :: Maybe String
   }
 
--- TODO: Could be a type class
-encodeStringifyArticle :: Article -> String
-encodeStringifyArticle = stringify <<< encodeJson <<< articleToJson
-
-encodeStringifyArticleStubs :: Array ArticleStub -> String
-encodeStringifyArticleStubs = stringify <<< encodeJson <<< map articleStubToJson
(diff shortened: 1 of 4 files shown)


packages/components/src/Lettera/Models.purs
```diff
@@ -225,12 +226,8 @@ type Author =
...
   }
 
--- TODO: Could be a type class
-encodeStringifyArticle :: Article -> String
-encodeStringifyArticle = stringify <<< encodeJson <<< articleToJson
-
-encodeStringifyArticleStubs :: Array ArticleStub -> String
-encodeStringifyArticleStubs = stringify <<< encodeJson <<< map articleStubToJson
+articleToArticleStub :: Article -> ArticleStub
+articleToArticleStub = unsafeCoerce
```
vapaj, line 230: Why `unsafeCoerce`? Wouldn't it lead to horrible bugs if we change the representation of either type?

kaol: Convenience, execution speed and the way Article and ArticleStub are implemented in Lettera we'd have to do something very special to have them diverge in other fields besides body.

But I changed it to have assignments, field by field. If `Article` and `ArticleStub` had all the fields defined it could be done with just `Record.delete (Proxy :: Proxy "body")` (which, under the hood, is just `unsafeCoerce`).

vapaj (approved):

PR #1200, opened 2022-06-07
Add Tailwind

jaakkoha-bytecraft:


apps/mosaico/src/Mosaico/Footer.purs
```diff
@@ -3,50 +3,60 @@ module Mosaico.Footer
...
 import React.Basic.DOM as DOM
-import React.Basic.Events (handler, EventHandler)
 import React.Basic.DOM.Events (preventDefault)
+import React.Basic.Events (handler, EventHandler)
 
 footer :: Paper -> (String -> EventHandler) -> JSX
 footer mosaicoPaper onStaticPageClick =
   DOM.footer
-    { className: "mosaico--footer"
+    { style: DOM.css {"gridArea": "foot"}
```
danielrlc, line 19: Nice you've used this kind of style too. I've found CSS Grid stuff that doesn't fit comfortably into Tailwind is best put in inline styles like this, to keep everything collocated. No need for a custom CSS class.

jaakkoha-bytecraft: Yeah, I'm not exactly sure what would be the best thing here... I did also consider writing a small TW plugin for the grid areas, since we are using those quite a lot. Maybe we should just move away from the CSS grid? I don't think it's the right piece of technology for Mosaico anyways.

danielrlc: Although how about using `[grid-area:foot]` instead?

apps/mosaico/tailwind.config.js
```diff
@@ -0,0 +1,109 @@
...
+      vn: "#c90c0f",
+      brand: "var(--brand-color)", // defined in _site.scss
+    },
+
+    fontFamily: {
+      roboto: ["Roboto", ...sansFonts],
+      duplexsans: ['"Duplex Sans"', ...sansFonts],
+      duplexserif: ['"Duplex Serif"', ...serifFonts],
+      mono: ["ui-monospace", ...monoFonts],
+    },
```
danielrlc, line 100: Sweet! ❤️

danielrlc (approved):
I left a few comments. Layout checked against `master` and all looks good to me.

apps/mosaico/src/Mosaico/Footer.purs
danielrlc: > Maybe we should just move away from the CSS grid? I don't think it's the right piece of technology for Mosaico anyways.
Pretty much agree. I'm not sure we need it.

danielrlc: And also, I've been a bit surprised by how invisible `gap` can end up. A useful concept I'm sure with certain kinds of layouts, but probably not for Mosaico. There's been a few times where I've been wondering why a white space was there, before realising it's because of gap on a parent.

less/mosaico/advertorial.less
```diff
@@ -28,7 +24,7 @@
 .advertorial-top-banner__company {
   align-self: center;
   width: 50%;
-  margin-left: @page-margin;
+  margin-left: 24px;
```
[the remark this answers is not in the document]
jaakkoha-bytecraft: Unfortunately no, I have only set up PostCSS (which handles TW) to only transform SCSS files.

PR #1201, opened 2022-06-07
Fall back to Mosaico list feed style if prerendered fails

kaol:


apps/mosaico/src/Main.purs
```diff
@@ -679,19 +678,27 @@ renderCategoryPage env (Category { label, type: categoryType, url}) = do
...
-  Cache.saveCategoryRender env.cache label $ mosaicoString $> rendered
-  now <- liftEffect nowDateTime
-  pure $ Cache.addHeader now mosaicoString $
-    htmlContent $ Response.ok $ StringBody rendered
+  case categoryType of
+    -- Fallback to using list feed style
+    Prerendered
+      | Just (Html html _) <- feed
+      , Right regex <- emptyRegex
+      , Just _ <- Regex.match regex html -> renderCategoryPage env $ Category $ category { type = Feed }
```
vapaj, line 686: How do we get an empty string as `html`?

kaol: It had a `fromMaybe ""` previously, but it wasn't that hard to bring it as a `Nothing` to that case since feed was a `Maybe` already.

I changed it but I still kept the test for empty string since it feels like a good sanity check to have.

vapaj (approved):

PR #1207, opened 2022-06-08
Mosaico/bruksvillkor redirect

danielrlc:


jaakkoha-bytecraft:
Should we have redirects for both? Just in case...

danielrlc:
Yes good point. I'll add them now.

danielrlc:
Added.

apps/mosaico/src/Main.purs
```diff
@@ -818,6 +818,9 @@ notFoundPage env {params: { path } } = do
     ["rss.xml"] /\ Paper.HBL -> redir "https://lettera.api.ksfmedia.fi/v4/list/frontpage?paper=HBL"
     ["rss.xml"] /\ Paper.VN -> redir "https://lettera.api.ksfmedia.fi/v4/list/frontpage?paper=VN"
     ["rss.xml"] /\ Paper.ON -> redir "https://lettera.api.ksfmedia.fi/v4/list/frontpage?paper=ON"
+    ["bruksvillkor/"] /\ Paper.HBL -> redir "https://www.hbl.fi/sida/bruksvillkor"
```
kaol, line 821: This won't work, it needs to be `["bruksvillkor",""]` (see a few lines above for an example).

danielrlc: Thanks. Fixed now.

kaol (approved):

PR #1225, opened 2022-06-16
Mosaico/abi subs

joakimekroos:
- Add static page for ABI (student) subscriptions.
- Add redirect for /abi -> /sida/abi


less/mosaico/static-pages/abi.less
```diff
@@ -0,0 +1,16 @@
+.static-page__abi {
```
MicciKarkia, line 1: The page overflows on mobile. There is a static-page.less file that can be used, especially if the font should be as the other static pages (using Duplex sans for headlines and Roboto for the rest of the things).

apps/mosaico/static/hbl/abi.html
```diff
@@ -0,0 +1,40 @@
+<div class="static-page__abi">
```
MicciKarkia, line 1: I think this file is missing Prettier formatting.

less/mosaico/static-pages/abi.less
```diff
@@ -0,0 +1,16 @@
...
+  padding: 60px 0;
+  .abi__headline {
```
MicciKarkia, line 3: If font-size is bigger than 1.5rem/24px, then some of the words in the headline are too long to fit on one line, so that might need some rules.

MicciKarkia (approved):

apps/mosaico/static/hbl/abi.html
joakimekroos: Now It's prettier

apps/mosaico/static/hbl/abi.html
```diff
@@ -0,0 +1,40 @@
...
+  <p>Prenumerationen börjar tre veckor innan studentprovet i svenska. I gåvoprenumerationen ingår
+  HBL:s elektroniska prenumeration i en månads tid.</p>
+  <p>Gåvoprenumerationen börjar 5.9.2022. För abiturienter är prenumerationen sålunda ett utmärkt
+  sätt att förbereda sig inför de kommande studentskrivningarna. Det finns ett begränsat antal
+  gåvoprenumerationer så det gäller att vara snabb. De första 2000 får prenumerationen HBL 365 för
+  en månad. HBL-appen kan laddas ner från App store (iOS), Google Play (Android) och Windows store
+  (Windows Phone).</p>
+  <p>Vi ber Er att snarast möjligt skicka in skolans uppgifter och – med elevernas tillåtelse –
+  abiturienternas adressuppgifter och e-post. Uppgifterna som behövs över studenterna är förnamn,
+  efternamn, gatuadress, postnummer, ort, telefonnummer, och e-post. Fyll i <a href="http://forms.ksfmedia.fi/fi/node/add/studentprenumeration">formuläret</a> senast
```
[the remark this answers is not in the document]
joakimekroos: I think the default language on that Drupal site is english, otherwise we could just disable it. I'll have a look.

PR #1233, opened 2022-06-17
Migrate profile component to use React hooks

kaol:


joakimekroos (approved):
Spun it up locally, seems to be working fine.

PR #1288, opened 2022-08-12
Only match first part of path

vapaj:
Ignore whatever comes after `path` given and do the redir. Eg `hbl.fi/abi`, `hbl.fi/abi/` and `hbl.fi/abi/asdasd/vdvsvnjdvsd` all redir to `hbl.fi/sida/abi`.


vapaj:
A better solution https://github.com/KSF-Media/affresco/pull/1292

PR #1289, opened 2022-08-16
Remove summer campaign

vapaj:


pikkuruusu (approved):
We should probably have some easy way to add these.

PR #1302, opened 2022-09-05
Add door code field to delivery reclamation redelivery form

kaol:


packages/components/src/DeliveryReclamation.purs
```diff
@@ -59,6 +61,7 @@ deliveryReclamation = make component { initialState, render, didMount }
 initialState :: State
 initialState =
   { publicationDate: Nothing
+  , doorCode: Just ""
```
vapaj, line 64: Why not `Nothing`?

kaol: `Nothing` means that the form would fail validation until the field got some value. I want it to succeed with initial empty value.

vapaj (approved):

PR #1309, opened 2022-09-27
Added deprecation warning to old app-article

pikkuruusu:
We want to shutdown the old app-article. But lets warn the few user we have first.


apps/app-article/src/App.js
```diff
@@ -27,6 +27,7 @@ import ManuallyRelatedArticles from "./components/manually-related-articles";
 import Cookies from "js-cookie";
 import { AndroidView } from "react-device-detect";
 import { Sentry } from "./sentry";
+import DepricationWarning from "./components/deprication-warning";
```
jaakkoha-bytecraft, line 30: sp: deprecation

pikkuruusu: Äsch, I figured that it could be wrong but I never bothered checking. 😅

jaakkoha-bytecraft (approved):

PR #1334, opened 2022-11-24
Add a limit parameter for fetching categories

jaakkoha-bytecraft:


jaakkoha-bytecraft:
wip while with mosaico's end as well

joakimekroos (approved):

PR #1336, opened 2022-12-02
Use time offset in time string instead of Z

kaol:


joakimekroos (approved):

packages/components/src/Helpers.purs
```diff
@@ -73,6 +77,55 @@ dateTimeFormatter =
...
+                         else "+" <> offsetStr offset
+    ]
+    where
+      padLeadingZero n | n < 10 = "0" <> show n
+      padLeadingZero n = show n
+      offsetStr o = padLeadingZero (div o 60) <> ":" <> padLeadingZero (rem 0 60)
+
+-- ISO 8601 parser.  Returns offset but doesn't apply it.
+parseLocalDateTime :: String -> Maybe { offset :: Int, dateTime :: DateTime }
+parseLocalDateTime str = do
```
jaakkoha-bytecraft, line 108: Is there a reason why this isn't just the purs equivalent of 
```js
function parseLocalDateTime(str) {
     const date = new Date(str);
    return date && { offset: date.getTimezoneOffset(), dateTime: date }
}
```

jaakkoha-bytecraft: ah right, because the offset is the input str's offset instead of the timezone offset. I'd still prefer not implementing a parser since the browser ships with one.

kaol: It's not going to be any easier to try to understand the whole of JS world's date handling for that.

jaakkoha-bytecraft (approved):

packages/components/src/Helpers.purs
jaakkoha-bytecraft: Yeah, misread a bit on what the code was actually doing, my bad.

PR #1372, opened 2023-02-20
Transform site article urls to app-article urls

pikkuruusu:


apps/app-article-server/src/browser/components/article-content.js
```diff
@@ -251,6 +259,25 @@ class Content extends Component {
...
+  handleInternalLinks(htmlString) {
+    const validHostnames = ['www.hbl.fi', 'www.ostnyland.fi', 'www.vastranyland.fi']
+
+    const $ = cheerio.load(htmlString, null, false)
+    const $links = $('a')
+    $links.each((i, el) => {
+      const $el = $(el)
+      const urlObj = new URL($el.attr('href'))
+
+      if(validHostnames.includes(urlObj.hostname) && urlObj.pathname.startsWith('/artikel/')) {
```
jaakkoha-bytecraft, line 271: Are the intra-article links always to www.hbl.fi and never to hbl.fi? The validHostnames could include the non-www alternatives to be sure.

apps/app-article-server/src/browser/components/article-content.js
```diff
@@ -251,6 +259,25 @@ class Content extends Component {
...
+        let linkUuid = urlObj.pathname.slice(9)
```
jaakkoha-bytecraft, line 272: And the 9 here could be replaced with "/artikel/".length or something similar, to show what it actually is removing.

apps/app-article-server/src/browser/components/article-content.js
pikkuruusu: Will probably be only to www, but yeah I don't think it hurts to add them just in case.

apps/app-article-server/src/browser/components/article-content.js
pikkuruusu: Good comment, added that.

jaakkoha-bytecraft (approved):

PR #1373, opened 2023-02-22
Add id's to elements in vetrina for triggerbee

MicciKarkia:


packages/vetrina/src/Vetrina/Purchase/NewPurchase.purs
```diff
@@ -174,8 +174,9 @@ title self =
   in foldMap headline headlineText
   where
     headline child =
-      DOM.h1
-        { className: "vetrina--headline-" <> maybe "KSF" Paper.toString self.props.paper <>
+      DOM.div
```
kaol, line 177: Just verifying: The change from h1 to div is intentional?

MicciKarkia: I would say so, because in mosaico we add a div element inside this element, so I thought that a div instead of a h1 would be more fitting

kaol (approved):

PR #1378, opened 2023-03-15
Remove ads based on setting in Lettera

danielrlc:


apps/app-article-server/src/browser/components/article.js
```diff
@@ -16,10 +16,12 @@ var _ = require("lodash");
 class Article extends Component {
   constructor(props) {
     super(props);
+    const adsAreShown = this.props.articleType !== "Advertorial" && !this.props.removeAds;
     this.state = {
       modalCaption: "",
       isImageModalOpen: false,
       mostReadArticles: [],
+      adsAreShown: adsAreShown,
```
kaol, line 24: I haven't tried running this code (and I'm still not a JS expert) so this might be covered by something I'm not seeing. But I see that this is setting `this.state.adsAreShown` and elsewhere in the code I'm seeing `this.props.adsAreShown` being used. Is everything correct with this?

pikkuruusu (approved):

PR #1385, opened 2023-04-03
Move Nets into a tab

jaakkoha-bytecraft:


packages/components/src/Window.purs
```diff
@@ -0,0 +1,9 @@
+module KSF.Window (close, clearOpener) where
+
+import Prelude
+
+import Effect (Effect)
+import Web.HTML.Window (Window)
+
+foreign import close :: Window -> Effect Unit
```
jaakkoha-bytecraft, line 8: There's `close` in the latest Web.HTML, but I tried upgrading to that and got some compile errors, so I wrapped it here rather than expanding this pull too much.

apps/mitt-konto/src/MittKonto/Main/Types.purs
```diff
@@ -23,6 +24,7 @@ type State =
   , now :: Date
   , news :: Maybe JSX
   , loginComponent :: Login.Props -> JSX
+  , window :: Maybe (Maybe Window)
```
kaol, line 27: A comment about how `Nothing` and `Just Nothing` differ would be ince.

kaol (commented):
I think the "Betalningen öppnas i ett nytt fönster" messages should all come with an avbryt button. It's poor UX if refreshing the screen is only way to cancel and the existing iframes offer an avbryt option inside the nets terminal.

apps/mitt-konto/src/MittKonto/Main/Types.purs
jaakkoha-bytecraft: > I think the "Betalningen öppnas i ett nytt fönster" messages should all come with an avbryt button.

I agree. I took a quick shot at this but cancelling the process from the modal is surprisingly convoluted (the same applies at the campaign plugin side). I'd just take the hit from the UX wart instead, as the other option seems to involve rewriting a lot of the handling code, just to get a cancel button.

apps/prenumerera/src/Page/Payment.purs
```diff
@@ -44,11 +50,25 @@ component :: Component Props
...
     orderState /\ setOrderState <- useState' $ Right OrderUnknownState
     netsUrl /\ setNetsUrl <- useState' Nothing
     error /\ setError <- useState' $ Right unit
     paperInvoiceConfirmed /\ setPaperInvoiceConfirmed <- useState' false
+    let needTerminal = method == CreditCard &&
+                       (error *> orderState) == Right OrderCreated &&
+                       isJust netsUrl
+        paymentUrl = (maybe "" _.paymentTerminalUrl netsUrl)
+    useEffect needTerminal do
+      if needTerminal then case window of
```
kaol, line 64: I think `when` would read better.

This may be code I wrote myself but I'll comment regardless.

packages/vetrina/src/Vetrina/Purchase/NewPurchase.purs
```diff
@@ -227,30 +231,43 @@ form self = DOM.form $
...
+    onSubmit = handler preventDefault $ \_ -> do
+      window <- Web.HTML.window
+      case self.props.accountStatus of
+        NewAccount ->
+          validation
+            (\_ ->
+              self.setState _
+                { newAccountForm
+                  { emailAddress = self.state.newAccountForm.emailAddress <|> Just "" }})
+            (\validForm -> do
```
kaol, line 243: I'm seeing this same lambda 3 times, let's use a `let`.

apps/prenumerera/src/Page/Payment.purs
jaakkoha-bytecraft: I think the original was yours, I just made it worse.

PR #1386, opened 2023-04-04
Improve search

joakimekroos:
Add `autoFocus` prop to the **inputField** module. Allows to set a field in focus when page renders.

### DIFF
--- a/packages/components/src/InputField/InputField.purs
+++ b/packages/components/src/InputField/InputField.purs
@@ -35,12 +35,14 @@ type Props =
   , disabled        :: Boolean
   , extraClass      :: String
   , autoComplete    :: String
+  , autoFocus       :: Boolean
   )
 
 type DefaultProps =
   ( disabled        :: Boolean
   , extraClass      :: String
   , autoComplete    :: String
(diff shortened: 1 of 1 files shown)


packages/components/src/InputField/InputField.purs
```diff
@@ -101,6 +104,7 @@ render self@{ props, state } =
                 else mempty
             , disabled: props.disabled
             , autoComplete: if props.autoComplete == "" then unsafeCoerce Nullable.null else props.autoComplete
+            , autoFocus: props.autoFocus
```
kaol, line 107: What are the exact semantics of setting this `false` instead of `null`? Initial Mosaico had a bunch of problems because it set all the Boolean fields as false and that didn't do what you'd expect with boolean attributes like `autoFocus`. Could you please check if this will end up making every input field have it enabled?

kaol: Fine by me. This records stuff is by necessity a bit low level and some coercing to `null`s may be required.

kaol (approved):

PR #1388, opened 2023-04-06
fix out-of-sync keys caused by inserted ads

danielrlc:
https://trello.com/c/Cg5scatM/1303-app-article-servers-frontend-rendering-doesnt-match-backends-rendering

Tested on the factbox in this article: http://localhost:8080/article/44f345b5-eb92-4e60-a884-26396cd3d831?paper=test  (test parameter makes test ads appear)

We've been using the array index to create keys for article items, but this has led to a mismatch between keys created before and after ads have been inserted into the article. So instead, as soon as the article has been fetched, the ad slots are inserted and a uuid-generated key is then created for each article item. The keys then don't change on render.

The factbox in the test article should open up, if the keys are in sync.

### DIFF
--- a/apps/app-article-server/src/browser/components/article-content.js
+++ b/apps/app-article-server/src/browser/components/article-content.js
@@ -11,8 +11,9 @@ class Content extends Component {
     };
   }
 
-  conditionalRendering(block, key) {
-    let blockType = Object.keys(block)[0].toLowerCase() || null;
+  conditionalRendering(block) {
+    const blockType = Object.keys(block)[0].toLowerCase() || null;
+    const key = block.key;
     if (blockType != null) {
       if (blockType === "ad") {
         return this.renderAd(block, key);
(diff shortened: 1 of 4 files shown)


apps/app-article-server/src/browser/components/article-content.js
```diff
@@ -211,8 +212,8 @@ class Content extends Component {
           <div
             className={`expand ${this.props.darkModeEnabled ? "darkMode" : ""}`}
             id={"expandFactBox-" + key}
-            onClick={(ev) => {
-              this.expandFactBox(ev);
+            onClick={() => {
+              this.expandFactBox(key);
             }}
```
danielrlc, line 217: I reverted this code to check if the old code worked. It's an open question whether we want to revert it back or not. (Applies to line 273 as well).

joakimekroos (approved):

PR #1392, opened 2023-04-17
Fix paywall styling to handle Triggerbee campaign

MicciKarkia:


kaol (approved):

less/Vetrina.less
```diff
@@ -98,6 +98,10 @@
     }
 }
 
+triggerbee-widget {
```
jaakkoha-bytecraft, line 101: I guess this should be `.triggerbee-widget` (dot in front)?

MicciKarkia: image
Well, it's an element called that, not an element with that class

jaakkoha-bytecraft: Ah, that's a weird tag name. Yeah, the code is correct then.

PR #1393, opened 2023-04-21
Tell Google which parts of our news are premium only

jaakkoha-bytecraft:


joakimekroos (approved):

packages/components/src/Lettera/ArticleSchema.purs
```diff
@@ -41,6 +48,13 @@ articleToJsonLd article =
             }
         )
         article.authors
+  , isAccessibleForFree: if article.premium then "False" else "True"
+  , hasPart: if article.premium
+             then Just { "@type": "WebPageElement"
+                       , "isAccessibleForFree": "False"
+                       , "cssSelector": ".premium-only"
```
kaol, line 55: This selector matches with no content on Mosaico. My reading of Google's instructions say that the selector has to match with the part of the document that's paywalled.

jaakkoha-bytecraft: https://git.ksfmedia.fi/taco/mosaico/-/merge_requests/233

PR #1394, opened 2023-04-21
Improve app article loading

pikkuruusu:
What this does:
• Uses React.lazy for the most read list component (don't think this do much when we return the page from the server)
• Lazy loads all images except main header image. Uses react-lazy-load-component for this.
• Use smaller file formats for fonts, woff2 seems to have been supported for quite a while now.
• We seemed to be loading some css twice. Stopped doing that.

### DIFF
--- a/apps/app-article-server/src/browser/components/article.js
+++ b/apps/app-article-server/src/browser/components/article.js
@@ -4,7 +4,6 @@ import Header from "./header";
 import Additional from "./article-additional";
 import ArticleDetails from "./article-details";
 import Content from "./article-content";
-// import MostReadArticles from "./most-read-articles";
 import Footer from "./footer";
 import RelatedArticles from "./related-articles";
 import Lightbox from "react-image-lightbox";
--- a/apps/app-article-server/src/browser/components/article.js
+++ b/apps/app-article-server/src/browser/components/article.js
@@ -1,14 +1,16 @@
-import React, { Component } from "react";
+import React, { Component, Suspense } from "react";
 import hblDefaultImage from "../assets/images/hbl-fallback-img.png";
 import Header from "./header";
 import Additional from "./article-additional";
 import ArticleDetails from "./article-details";
 import Content from "./article-content";
-import MostReadArticles from "./most-read-articles";
+// import MostReadArticles from "./most-read-articles";
 import Footer from "./footer";
 import RelatedArticles from "./related-articles";
 import Lightbox from "react-image-lightbox";
 import "react-image-lightbox/style.css";
+
+const MostReadArticles = React.lazy(() => import('./most-read-articles'))
 
 const axios = require("axios");
 var _ = require("lodash");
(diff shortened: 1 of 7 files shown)


apps/app-article-server/src/browser/components/article.js
```diff
@@ -1,15 +1,17 @@
-import React, { Component } from "react";
+import React, { Component, Suspense } from "react";
 import hblDefaultImage from "../assets/images/hbl-fallback-img.png";
 import Header from "./header";
 import Additional from "./article-additional";
 import ArticleDetails from "./article-details";
 import Content from "./article-content";
-import MostReadArticles from "./most-read-articles";
+// import MostReadArticles from "./most-read-articles";
```
jllang, line 7: Do we need this comment?

apps/app-article-server/src/browser/index.css
```diff
@@ -1,21 +1,27 @@
 @font-face {
   font-family: "Roboto";
-  src: url("https://cdn.ksfmedia.fi/assets/fonts/Roboto-Regular.ttf");
+  src: url("https://cdn.ksfmedia.fi/assets/fonts/Roboto-Regular-webfont.woff2") format("woff2");
```
jaakkoha-bytecraft, line 3: We could also include both of the fonts here and let the browser decide.

The format is something around:
``` 
src:  url('Montserrat-Regular.woff') format('woff'), 
         url('Montserrat-Regular.ttf')  format('truetype'),
         url('Montserrat-Regular.svg#Montserrat') format('svg');
```

apps/app-article-server/src/browser/components/mobile-article-list.js
```diff
@@ -64,14 +65,18 @@ const MobileList = (props) => {
...
+                    <>
+                    <LazyLoadImage
                       src={item.listImage.url.includes("imengine")
-                          ? item.listImage.url + "&function=hardcrop&width=360&height=200&q=65"
-                          : item.listImage.tinyThumb
-                          }
-                      alt=""
+                      ? item.listImage.url + "&function=hardcrop&width=360&height=200&q=65"
+                      : item.listImage.tinyThumb
+                      }
```
jllang, line 73: I don't know what kind of JS style convention we're following, but my intuition tells to use indentation here.

jllang (approved):

apps/app-article-server/src/browser/components/mobile-article-list.js
jaakkoha-bytecraft: `npx prettier --write src/browser/components/mobile-article-list.js` formats stuff nicely.

apps/app-article-server/src/browser/index.css
pikkuruusu: True, I'll add that.

apps/app-article-server/src/browser/components/mobile-article-list.js
pikkuruusu: Good catch, I'll fix it.

PR #1402, opened 2023-04-27
Advertorial liftups for app-article-server

jllang:
Add advertorial liftups to articles with ads.

Staging Lettera doesn't seem to have active advertorials. Set `LETTERA_V4_URL=https://lettera.api.ksfmedia.fi/v4` in `.env` file (but don't commit it) in order to be able to test how advertorial liftups work.

### DIFF
--- a/apps/app-article-server/.env.local
+++ b/apps/app-article-server/.env.local
@@ -1,5 +1,6 @@
 PERSONA_URL=https://persona.staging.ksfmedia.fi/v1
 LETTERA_URL=https://lettera.staging.ksfmedia.fi/v3
+LETTERA_V4_URL=https://lettera.staging.ksfmedia.fi/v4
 SHARE_URL='https://app-article.ksfmedia.fi/index.html?uuid='
 SENTRY_DSN=
 PUBLIC_URL=/dist
--- a/apps/app-article-server/app.dev.yaml
+++ b/apps/app-article-server/app.dev.yaml
@@ -1,6 +1,0 @@
-runtime: nodejs12
-service: app-article-server
-entrypoint: node -r dotenv/config dist/server
-env_variables:
-  PERSONA_URL: "https://persona.api.ksfmedia.fi/v1"
-  LETTERA_URL: "https://lettera.api.ksfmedia.fi/v3"
--- a/apps/app-article-server/src/browser/components/advertorial-liftup.js
+++ b/apps/app-article-server/src/browser/components/advertorial-liftup.js
@@ -1,30 +1,37 @@
-import React from "react";
+import React, { useEffect, useState } from "react";
+const axios = require("axios");
 
-const AdvertorialLiftup = (props) => {
+export function AdvertorialLiftup({ darkModeEnabled, paper }) {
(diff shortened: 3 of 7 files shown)


apps/app-article-server/app.dev.yaml
jllang, line 1: @mrskug requested removing this file

apps/app-article-server/src/browser/components/article.js
```diff
@@ -31,6 +32,7 @@ class Article extends Component {
       document.getElementsByTagName("HTML")[0].setAttribute("data-theme", "dark");
     }
     this.getMostReadArticles();
+    this.getAdvertorialLiftup();
```
jllang, line 35: Is this the correct way of implementing lazy loading?

danielrlc: Looks like you need to import it lazily, the same way as with MostReadArticles on line 13.

apps/app-article-server/src/browser/components/advertorial-liftup.js
```diff
@@ -0,0 +1,30 @@
...
+        />
+      </div>
+      <h2 className="mt-3 text-3xl font-semibold break-words font-robotoslab">
+        {props.listTitle}
+      </h2>
+    </a>
+  );
+};
+
+export default AdvertorialLiftup;
```
jaakkoha-bytecraft, line 30: I personally recommend avoiding default exports when working with JSX/TSX: it makes IDE autocomplete less effective. With a named export (`export const AdvertorialLiftup = ...`) you get better autocomplete suggestions when writing code. It also helps maintaining constant naming across the codebase, as you can pretty easily have different names for any default export.

jaakkoha-bytecraft: That said, React.lazy only works with default exports, so for this specific case it might be easier to stick to the default export. The other option would be to use eg. `const AdvertorialLiftup = React.lazy(()=> import('./advertorial-liftup.js').then(module=>({default:module.AdvertorialLiftup})))`.

apps/app-article-server/src/browser/components/article.js
pikkuruusu: Actually I would say it would be best to use `react-lazy-load-image-component` library the same way it is used in article-content for the image the advertorial-liftup-component. So:

`import { LazyLoadImage } from "react-lazy-load-image-component";`

and 

```
<LazyLoadImage 
  className={"adv-image"}
  src={props.image}
  threshold="200"
  effect="opacity"
  alt=""
/>
 ```
 
 I'm not sure React.lazy does much when it's returned from the server.

File: apps/app-article-server/.env.local [hunk 3]
```diff
@@ -1,5 +1,6 @@
 PERSONA_URL=https://persona.staging.ksfmedia.fi/v1
 LETTERA_URL=https://lettera.staging.ksfmedia.fi/v3
+LETTERA_V4_URL=https://lettera.staging.ksfmedia.fi/v4
```
jllang, line 3: TODO: Add this to infra

pikkuruusu (approved):

PR #1407, opened 2023-05-15
Eslint and prettier

jaakkoha-bytecraft:


jaakkoha-bytecraft:
Most of this PR is just line noise from `prettier`, I guess I'll create another PR for an git blame ignore file, once the merge hash is known.

apps/app-article-server/src/browser/components/article-content.js
```diff
@@ -20,9 +20,7 @@ class Content extends Component {
       if (blockType === "ad") {
         return this.renderAd(block, key);
       }
-      if (blockType === "paragraph") {
-        return this.renderParagraph(block, key);
-      } else if (blockType === "headline") {
```
jaakkoha-bytecraft, line 25: One of the very few non-formatter changes: `htmlToReactParser` is always `undefined` in the `renderParagraph` function, so calling `renderParagraph` would always fail -- since it hasn't been a problem, we aren't rendering any blocks with the type being `paragraph`.

apps/mitt-konto/index.js
```diff
@@ -2,15 +2,10 @@ import React from "react";
 import ReactDOM from "react-dom";
 import "basscss/css/basscss-cp.css";
 
-// yup, welcome to react 16
-import createReactClass from "create-react-class";
-React.createClass = createReactClass;
```
jaakkoha-bytecraft, line 7: A handful of services define `createClass` in their index.js files. None of the generated code refers to that function.

PR #1413, opened 2023-05-25
Nets soft fails

jaakkoha-bytecraft:


packages/vetrina/src/Vetrina/Vetrina.purs
```diff
@@ -638,3 +652,31 @@ netsTerminalModal =
           ]
       ]
     }
+
+scaRequired :: Maybe String -> (Boolean -> Effect Unit) -> JSX
```
kaol, line 656: I'll approve this but this `Maybe` is still pretty annoying.

kaol (approved):

PR #1428, opened 2023-06-14
Migrate app-article-server to Lettera V4

jllang:


jllang:
We need to remember to reconfigure the `LETTERA_URL` variable for staging and production to use Lettera V4 API when we deploy this migration.

mrskug (approved):

PR #1431, opened 2023-06-15
Ospedale/hello world

kaol:


apps/ospedale/src/Ospedale.purs
```diff
@@ -0,0 +1,148 @@
...
+          { paper: fromMaybe Paper.KSF paper
+          , activeUser: _.email <$> session
+          , specialHelp: Nothing
+          , logout
+          }
+      , case {loading, session} of
+            {loading: Just _} -> Spinner.loadingSpinner
+            {session: Nothing} -> loginComponent {setSession: setSession <<< Just}
+            {session: Just s} -> renderSite (Just s.session)
+      ]
```
jllang, line 132: This function is almost 100 lines. Would it make sense to factor it into smaller pieces?

apps/ospedale/src/Ospedale.purs
```diff
@@ -0,0 +1,148 @@
...
+
+jsApp :: {} -> JSX
+jsApp = unsafePerformEffect app
```
jllang, line 135: I don't know about PureScript, but I think that `unsafePerformIO` is generally regarded as an antipattern in Haskell.

apps/ospedale/src/Page/Edit.purs
```diff
@@ -0,0 +1,283 @@
...
+          f x) $ useState' Nothing
+    error /\ setError <- useState' Nothing
+
+    let submit token = do
+          now <- Now.nowDateTime
+          nowLocal <- fromUTCTime now
+          let publishingTime :: Maybe LocalDateTime
+              publishingTime =
+                if not publish then Nothing
+                else (_.publishingTime =<< inputArticle) <|> nowLocal
```
jllang, line 76: We could use guards instead of `if` here.

apps/ospedale/src/Page/Edit.purs
```diff
@@ -0,0 +1,283 @@
...
+                          (if isJust (_.publishingTime <$> savedArticle)
+                           then " och publicerat" else "") <>
+                          (if modified then ". Artikeln har ändrats efter att ha sparats." else "")
+                        ]
+            }
+      ]
+  where
+    -- Assume only HTML elements
+    buildBody (Html html) = html
+    buildBody _ = ""
```
jllang, line 259: Again, we have quite a lot of code. Personally, I'd prefer smaller pieces.

PR #1435, opened 2023-06-20
First attempt at creating a paywall disabling button

jllang:
Add a new page for managing paywall openings

Kuvakaappaus2023-07-13 11-08-44

### DIFF
--- a/apps/mitt-konto/src/Components/Paywall.purs
+++ b/apps/mitt-konto/src/Components/Paywall.purs
@@ -2,7 +2,7 @@ module MittKonto.Components.Paywall where
 
 import Prelude
 
-import Data.Array ((:), filter, findIndex, head, modifyAt, null)
+import Data.Array ((:), filter, findIndex, modifyAt, null)
 import Data.Int (fromString)
 import Data.Maybe (Maybe(..), fromMaybe, maybe)
 import Effect (Effect)
@@ -10,8 +10,7 @@ import Effect.Aff as Aff
 import Effect.Class (liftEffect)
 import Effect.Exception (error)
 import Effect.Uncurried (mkEffectFn1)
-import Data.JSDate (JSDate, now, toUTCString)
-import KSF.Api.Entitlements (PaywallOpening(..))
+import KSF.Api.Entitlements (PaywallOpening)
 import KSF.Sentry as Sentry
 import KSF.Spinner (loadingSpinner)
 import KSF.User as User
@@ -43,7 +42,6 @@ initialProducts =
 
 paywall :: PushStateInterface -> Sentry.Logger -> Component Props
 paywall _router _logger = do
-  currentDate <- now
   component "Paywall" \ {} -> React.do
     days     /\ setDays     <- useState' 0
     hours    /\ setHours    <- useState' 0
--- a/apps/mitt-konto/src/Components/Paywall.purs
+++ b/apps/mitt-konto/src/Components/Paywall.purs
@@ -77,7 +77,7 @@ paywall _router _logger = do
             }
         , DOM.hr {}
         , DOM.h2_ [ DOM.text "Nuvarande betalmur öppningar" ]
-        , maybe loadingSpinner renderCurrentOpenings openings
+        , maybe loadingSpinner (renderCurrentOpenings setEra) openings
         ]
       }
 
@@ -167,8 +167,11 @@ renderProducts products setProducts =
           }
       ]
 
-renderCurrentOpenings :: Array PaywallOpening -> JSX
-renderCurrentOpenings openings =
+renderCurrentOpenings
+  :: ((Int -> Int) -> Effect Unit)
+  -> Array PaywallOpening
+  -> JSX
+renderCurrentOpenings setEra openings =
   let
     renderOpening opening = DOM.tr
       { children:
@@ -176,7 +179,14 @@ renderCurrentOpenings openings =
           , DOM.td_ [ DOM.text (toUTCString opening.startAt) ]
           , DOM.td_ [ DOM.text (toUTCString opening.endAt) ]
           , DOM.td_ [ DOM.text (show opening.onlyToProducts) ]
-          , DOM.td_ [ DOM.text "radera" ]
+          , DOM.td_
+              [ DOM.button
+                  { children: [DOM.text "radera"]
+                  , onClick: mkEffectFn1 $ \_ -> Aff.launchAff_ $ do
+                      User.deletePaywallOpening 59
+                      liftEffect $ setEra (_ + 1)
+                  }
+              ]
           ]
       }
   in
--- a/apps/mitt-konto/src/Components/Paywall.purs
+++ b/apps/mitt-konto/src/Components/Paywall.purs
@@ -2,7 +2,7 @@ module MittKonto.Components.Paywall where
 
 import Prelude
 
-import Data.Array (filter, findIndex, modifyAt, null)
+import Data.Array ((:), filter, findIndex, head, modifyAt, null)
 import Data.Int (fromString)
 import Data.Maybe (Maybe(..), fromMaybe, maybe)
 import Effect (Effect)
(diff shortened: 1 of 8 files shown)


apps/mitt-konto/src/Components/Paywall.purs
```diff
@@ -0,0 +1,217 @@
...
+                  , checked: selected
+                  , onChange: mkEffectFn1 (\_ -> setProducts (toggleProduct name))
+                  }
+              , DOM.text label
+              ]
+          }
+      ]
+
+renderCurrentOpenings
+  :: ((Int -> Int) -> Effect Unit)
```
kaol, line 173: I would suggest changing this to `Int -> EventHandler` where `Int` is the id of the line and have the function definition on lines 189-191 be inside the let block starting on line 51.

packages/components/src/Persona.purs
```diff
@@ -7,7 +7,7 @@ import Data.Array (mapMaybe, catMaybes, filter)
 import Data.Date (Date)
 import Data.Date as Date
 import Data.Generic.Rep (class Generic)
-import Data.JSDate (JSDate, toDate)
+import Data.JSDate (JSDate, jsdate, toDate)
```
kaol: Unused import `jsdate`.

packages/components/src/Api/Entitlements.purs
```diff
@@ -0,0 +1,25 @@
+module KSF.Api.Entitlements where
+
+import Prelude
+import Data.JSDate (JSDate)
+import Data.Maybe (Maybe(..))
+import Data.Nullable (Nullable)
+import Foreign (Foreign, ForeignError(..), fail)
+import Foreign.Object (lookup)
+import Simple.JSON (class ReadForeign, readImpl)
```
kaol, line 9: Check imports, there are extra since you removed the `ReadForeign` instance.

apps/mitt-konto/src/Components/Paywall.purs
```diff
@@ -0,0 +1,217 @@
...
+      | n < 1     = Just 0
+      | n > max   = Just max
+      | otherwise = Just n
+
+renderDays :: Int -> (Int -> Effect Unit) -> JSX
+renderDays days setDays =
+  DOM.label
+    { children:
+        [ DOM.input
+            { type: "text"
```
kaol, line 99: Why not `type: "number"`?
Same for the other time fields.

apps/mitt-konto/src/MittKonto/Main/Views.purs
```diff
@@ -14,11 +14,13 @@ import Data.Monoid (guard)
 import Data.Nullable (toMaybe)
 import Data.String as String
 import Effect (Effect)
+import Effect.Uncurried (mkEffectFn1)
```
kaol: Unused.

apps/mitt-konto/src/Components/Paywall.purs
```diff
@@ -0,0 +1,217 @@
...
+import Prelude
+
+import Data.Array ((:), filter, findIndex, head, modifyAt, null)
+import Data.Int (fromString)
+import Data.Maybe (Maybe(..), fromMaybe, maybe)
+import Effect (Effect)
+import Effect.Aff as Aff
+import Effect.Class (liftEffect)
+import Effect.Exception (error)
+import Effect.Uncurried (mkEffectFn1)
```
kaol, line 12: Using `mkEffectFn1` feels a bit too low level to me, I think I'd prefer using `handler_` and friends from `React.Basic.Events`.

apps/mitt-konto/src/Components/Paywall.purs
```diff
@@ -0,0 +1,217 @@
...
+      fiber <- Aff.launchAff $
+        liftEffect <<< setOpenings <<< Just =<< User.getPaywallOpenings
+      pure $ Aff.launchAff_ $ Aff.killFiber (error "component closed") fiber
+    let
+      selection = map (\p -> p.name) $ filter (\p -> p.selected) products
+      submitHandler :: EventHandler
+      submitHandler = handler_ $ Aff.launchAff_ $ do
+        User.openPaywall days hours minutes selection
+        liftEffect $ setEra (_ + 1)
+    pure $ DOM.div
```
kaol, line 62: This could also use `DOM.form` at this level instead and use `onSubmit` to do what the `Skicka` button's `onClick` does. And use `type: "submit"` for that button, too.

apps/mitt-konto/src/MittKonto/Main/Views.purs
```diff
@@ -14,11 +14,13 @@ import Data.Monoid (guard)
...
 import Foreign (unsafeToForeign)
 import KSF.Alert.Component (Alert)
 import KSF.Alert.Component as Alert
 import KSF.Footer.Component as Footer
 import KSF.Navbar.Component as Navbar
+import KSF.User as User
```
kaol: Unused import.

kaol (changes requested):

apps/mitt-konto/src/Components/Paywall.purs
jllang: Using `DOM.input` with `type: "submit"` leads into page reload. The `type` attribute doesn't seem to be available for `DOM.button` on the other hand.

jllang: Using a `DOM.form` element seems to lead into page reload after each click of a button.

packages/user/src/User/User.purs
```diff
@@ -758,3 +764,28 @@ callBottega f = do
 
 getPackages :: Aff (Array Package)
 getPackages = Bottega.getPackages
+
+getPaywallOpenings :: Aff (Array PaywallOpening)
+getPaywallOpenings =
+  Persona.getPaywallOpenings =<< requireToken
+
+openPaywall :: Int -> Int -> Int -> Array String -> Aff Unit
+openPaywall days hours minutes products = do
```
kaol, line 773: You could use `onlyToProducts` as variable name instead of `products` and then you could use the `{onlyToProducts}` syntax on line 785.

packages/components/src/Persona.purs
```diff
@@ -87,6 +89,18 @@ getUserEntitlements :: UserAuth -> Aff (Array String)
 getUserEntitlements auth =
   callApi usersApi "usersUuidEntitlementGet" [ unsafeToForeign auth.userId ] $ authHeaders auth.userId auth
 
+getPaywallOpenings :: UserAuth -> Aff (Array PaywallOpening)
+getPaywallOpenings auth =
+  callApi entitlementsApi "entitlementsAllowGet" [] $ authHeaders UUID.emptyUUID auth
+
+openPaywall :: UserAuth -> AllowEntitlementsQuery -> Aff Unit
```
kaol, line 96: The style in this file has been to always have `UserAuth` as the last parameter, this could be changed to match with that.

kaol (changes requested):

packages/components/src/Api/Entitlements.purs
jllang: Should I get compiler warnings from unused imports? I don't.

apps/mitt-konto/src/Components/Paywall.purs
```diff
@@ -0,0 +1,226 @@
...
+      }
+
+setNumber :: Int -> (Int -> Effect Unit) -> EventHandler
+setNumber max setter = handler targetValue \value ->
+  maybe (pure unit) setter (value >>= handleEmpty >>= fromString >>= handleBounds)
+  where
+    handleEmpty s = Just case s of
+      "" -> "0"
+      _  -> s
+    handleBounds n
```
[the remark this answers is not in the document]
jllang: If I use the `min` and `max` attributes instead, I can't use the increment buttons to select too big values but I can type arbitrarily large numbers with keyboard and then I get a popup saying the input was too large:

Kuvakaappaus2023-07-13 09-05-13

I think it's better to go with a "correct by construction" approach that prevents ever entering incorrect values in the first place. But now that I think of it, maybe setting 70 minutes could translate into 1 h 10 min and 30 hours could be 1 d 6 h.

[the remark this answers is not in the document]
jllang: By the way, if the input type is `number`, then I can't use backspace to set it back to zero.

packages/components/src/Api/Entitlements.purs
jllang: I found the unused imports and removed them from this module.

kaol (approved):

PR #1444, opened 2023-07-14
Make it possible to open paywall in future

jllang:
Relates to https://trello.com/c/kgxOpoxg/1542-ability-to-add-future-paywall-openings-for-mitt-konto-admins

Kuvakaappaus2023-07-14 12-32-13

### DIFF
--- a/apps/mitt-konto/src/Components/Paywall.purs
+++ b/apps/mitt-konto/src/Components/Paywall.purs
@@ -2,16 +2,17 @@ module MittKonto.Components.Paywall where
 
 import Prelude
 
-import Data.Array ((:), filter, findIndex, length, modifyAt, null)
+import Data.Array ((:), filter, findIndex, length, modifyAt)
 import Data.Date (Date)
 import Data.DateTime (DateTime(..), date, time)
 import Data.Enum (class BoundedEnum, fromEnum, toEnum)
 import Data.Foldable (intercalate)
-import Data.Int (fromString, toNumber)
+import Data.Formatter.DateTime (FormatterCommand(..), format)
+import Data.Int (fromString)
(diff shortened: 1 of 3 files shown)


apps/mitt-konto/src/Components/Paywall.purs
```diff
@@ -23,6 +31,8 @@ import React.Basic.Hooks (Component, component, useEffect, useState, useState',
 import React.Basic.Hooks as React
 import Routing.PushState (PushStateInterface)
 
+type Setter a = a -> Effect Unit
```
kaol, line 34: Just write out the types, it's easier to read than custom wrappers like these, especially for this simple types like these.

kaol (changes requested):

jllang:
We decided in Slack to replace date, time, and duration inputs with two standard date selectors. The tradeoff is that the locale is determined by the user's browser. On my en-US firefox, the page now looks like

Kuvakaappaus2023-07-14 15-37-48

On my fi-FI Chromium, the page looks like this (notice also the style differences):
Kuvakaappaus2023-07-14 15-41-02

apps/mitt-konto/src/Components/Paywall.purs
```diff
@@ -151,11 +147,8 @@ toggleProduct name products =
     let modify p = p { selected = not p.selected }
     modifyAt index modify products
 
-renderProducts
-  :: Products
-  -> ((Products -> Products) -> Effect Unit)
-  -> JSX
-renderProducts products setProducts =
+renderProducts :: Products -> Modifier Products -> JSX
```
kaol, line 150: I'd rather read the function type without this alias.

kaol (approved):

PR #1452, opened 2023-07-25
Add support for removing child elements to the HTML renderer

jllang:


jllang:
I don't think we need this. I found another way to do what I needed to do.

PR #1453, opened 2023-07-26
Add utilities to KSF.Helpers

oneb:
Will be used by Mosaico to implement redirecting of /artikel/<editoriald> to published article.

### DIFF
--- a/packages/components/src/Helpers.purs
+++ b/packages/components/src/Helpers.purs
@@ -6,19 +6,24 @@ import           Control.Alt ((<|>))
 import           Data.Date               (Date, adjust)
 import           Data.DateTime           (DateTime (..))
 import           Data.DateTime           as DateTime
+import           Data.Either             (hush)
 import           Data.Enum               (toEnum)
 import           Data.Formatter.DateTime (Formatter, FormatterCommand (..),
                                           format)
 import           Data.Int                (toNumber)
 import           Data.List               (fromFoldable)
-import           Data.Maybe              (Maybe (..), fromJust)
+import           Data.Maybe              (Maybe (..), fromJust, isJust)
 import           Data.String             (Pattern (..))
 import           Data.String             as String
+import           Data.String.Regex       as Regex
 import           Data.Time               (Time (..))
 import           Data.Time.Duration      (Days (..), Hours (..), negateDuration)
+import           Data.UUID               as UUID
 import           Effect                  (Effect)
 import           Effect.Now              as Now
 import           Partial.Unsafe          (unsafePartial)
+
+
 
 foreign import getCurrentTZOffset_ :: Effect Int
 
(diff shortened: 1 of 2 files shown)


jllang (approved):

packages/components/src/Helpers.purs
```diff
@@ -117,3 +122,17 @@ getMinStartDate nextDelivery = do
...
+matchesRegex regex str = isJust $ do
+  re <- hush $ Regex.regex regex mempty
+  Regex.match re str
+
+-- This same deterministic function is in Lettera's code
+editorialIdToUuid :: String -> UUID.UUID
+editorialIdToUuid editorialId =
+  UUID.genv5UUID ("https://hblmedia.fi/" <> editorialId) url_namespace
+
+-- Constant defined in the UUIDv5 standard
```
kaol, line 136: Add a link as a reference to where this is defined

packages/components/src/Helpers.purs
```diff
@@ -117,3 +122,17 @@ getMinStartDate nextDelivery = do
...
+url_namespace :: UUID.UUID
+url_namespace = unsafePartial $ fromJust $ UUID.parseUUID "6ba7b811-9dad-11d1-80b4-00c04fd430c8"
```
kaol, line 138: Please set your editor to add line feed after the final line.

packages/components/src/Helpers.purs
```diff
@@ -117,3 +122,17 @@ getMinStartDate nextDelivery = do
           soonestStart = adjust soonestDuration (DateTime.date now)
           byNextIssue = max <$> soonestStart <*> nextDelivery
       pure $ byNextIssue <|> soonestStart
+
+matchesRegex :: String -> String -> Boolean
```
kaol, line 126: I can't say that I'm a fan of this function. Just using Regex directly is not that much different, this doesn't add that much. And this constructs a new Regex object for each call as well, presumably the use case is to use the same regex every time?

packages/components/src/Helpers.purs
```diff
@@ -117,3 +122,17 @@ getMinStartDate nextDelivery = do
...
+matchesRegex regex str = isJust $ do
+  re <- hush $ Regex.regex regex mempty
+  Regex.match re str
+
+-- This same deterministic function is in Lettera's code
+editorialIdToUuid :: String -> UUID.UUID
```
kaol, line 132: Move this to `Lettera` or `Lettera.Models` since this is specific to it.

packages/components/src/Helpers.purs
```diff
@@ -6,20 +6,25 @@ import           Control.Alt ((<|>))
...
 import           Data.String             as String
+import           Data.String.Regex       as Regex
 import           Data.Time               (Time (..))
 import           Data.Time.Duration      (Days (..), Hours (..), negateDuration)
+import           Data.UUID               as UUID
 import           Effect                  (Effect)
 import           Effect.Now              as Now
 import           Partial.Unsafe          (unsafePartial)
 
+
```
kaol, line 26: Extra space.

kaol (changes requested):

packages/components/src/Helpers.purs
oneb: The advantage of this is that it's concise and obvious, while the functions in the Regex module are a bit fiddly. But fine, it can be removed at the cost of a few extra lines and less clarity at the place where it was used. And then we can also precompile the regex object.

packages/components/src/Helpers.purs
oneb: Fixed

packages/components/src/Helpers.purs
oneb: All changes to this file undone, so fixed.

packages/components/src/Helpers.purs
oneb: Fixed

packages/components/src/Helpers.purs
oneb: Fixed

packages/components/src/Helpers.purs
oneb: Removed

kaol (approved):

PR #1454, opened 2023-07-26
Add DFS helper to Html render module

jllang:


packages/components/src/HtmlRenderer/Models.purs
```diff
@@ -125,3 +126,12 @@ toJSGenericHook h =
   , preprocessNode:       toNullable $ mkEffectFn2 <$> h.preprocessNode
   , processNode:          toNullable $ mkEffectFn3 <$> h.processNode
   }
+
+-- | Performs a Depth-First Search (DFS) on the given node and returns `true`
+--   once a children that satisfies the given predicate is found or `false`
+--   once all reachable nodes have been visited and none of them satisfied the
+--   predicate.
+dfs :: (Node -> Boolean) -> Node -> Boolean
+dfs prop node
```
kaol, line 135: I'd find it simpler to just have
```
dfs prop node = prop node || maybe false (any (dfs prop)) (getChildren node)
```

kaol (changes requested):

kaol (approved):

PR #1469, opened 2023-08-30
Add inputfield class to props and places using it

MicciKarkia:


packages/components/src/InputField/InputField.purs
```diff
@@ -101,7 +102,7 @@ render self@{ props, state } =
             , className:
                 if isJust props.validationError
                 then "input-field--invalid-field"
```
kaol, line 104: Doing it this way will remove all element specific classes in case there's a validation error.

kaol (changes requested):

packages/components/src/InputField/InputField.purs
MicciKarkia: I fixed it per your suggestion :)

kaol (approved):

PR #1473, opened 2023-09-04
Vetrina/use tailwind

MicciKarkia:


packages/vetrina/index.js
```diff
@@ -1,4 +1,4 @@
-require("./build/Vetrina.css");
```
joakimekroos, line 1: Delete rather than commenting out?

packages/vetrina/package.json
```diff
@@ -4,7 +4,7 @@
   "description": "Vetrina, frontend for purchasing Hufvudstadsbladet Ab products",
   "license": "MIT",
   "scripts": {
-    "build-npm": "spago bundle-module -m KSF.Vetrina -t build.js && parcel build --no-source-maps -d build ../../less/Vetrina.less",
```
joakimekroos, line 7: Unintentional extra space?

packages/vetrina/src/Vetrina/Products/Premium.purs
```diff
@@ -13,13 +13,13 @@ hblPremium =
   , description:
       DOM.div
         { id: "tb-paywall--product-text-hbl"
-        , className: "vetrina--product-text text-center text-base leading-tight px-5 pb-5"
```
joakimekroos, line 16: Should we remove the `vetrina--something` classes also from the DOM since most (all?) of them have been axed from Vetrina.less. I'm thining it can get confusing later on when someone, like "Why is this class here, where are its rules defined?"

joakimekroos (changes requested):

packages/vetrina/index.js
MicciKarkia: I actually think it's supposed to be there, I commented it out when I thought that we could altogether delete Vetrina less file, but then as it is needed for dark mode, I guess this should not be commented out or deleted

packages/vetrina/package.json
MicciKarkia: Yes! Thank you!

packages/vetrina/src/Vetrina/Products/Premium.purs
MicciKarkia: Yes, we talked about that but then if we only have the Tailwind classes in the DOM, then it makes debugging really hard. Or at least I haven't found an easy way for that..

joakimekroos (approved):

PR #1486, opened 2023-09-13
Feature/improve lettera error messages

jllang:


packages/components/src/Lettera.purs
```diff
@@ -76,15 +80,57 @@ letteraSearchUrl = letteraBaseUrl <> "/list/search"
...
+encodeError name req err =
+  "error" := name ~>
+  (("request" := _) <$> req) ~>?
+  (("message" := _) <$> err) ~>?
+  jsonEmptyObject
+
 data LetteraError
-  = ResponseError AX.Error
-  | HttpError Int
+  = ResponseError String AX.Error
```
kaol, line 126: I don't like this. If we want to do anything besides just print out the error (like retry with different parameters) this'd require parsing a string. It's better to leave it as `AX.Request` and not try to do anything else at this level.
Move the stringification to the `Show` instance.

packages/components/src/Lettera.purs
```diff
@@ -76,15 +80,57 @@ letteraSearchUrl = letteraBaseUrl <> "/list/search"
...
+  | HttpError String String
```
kaol, line 127: Status code is used in downstream, not just for logging, so it needs to stay as `Int`.

packages/components/src/Lettera.purs
```diff
@@ -76,15 +80,57 @@ letteraSearchUrl = letteraBaseUrl <> "/list/search"
 letteraAdvertorialUrl :: String
 letteraAdvertorialUrl = letteraBaseUrl <> "/list/active-advertorial"
 
+encodeRequestHeaders :: Array AX.RequestHeader -> Maybe Json
```
kaol, line 83: `Json` should only be used for serialization/deserialization, not this.

packages/components/src/Lettera.purs
```diff
@@ -218,7 +269,7 @@ useResponse f (Right response)
                                     }
       Left err -> LetteraResponse { maxAge: Nothing, body: Left err }
   | otherwise =
-      pure $ LetteraResponse { maxAge: Nothing, body: Left $ HttpError $ unwrap $ response.status }
+      pure $ LetteraResponse { maxAge: Nothing, body: Left $ HttpError request $ stringify (encodeResponse response) }
```
kaol, line 272: `stringify` and `encodeResponse` are only needed because `Json` was used as the type. Remove them.

kaol (changes requested):

packages/components/src/Lettera.purs
jllang: I tried to have `AX.Request` there, but I got  type problems. I'll try to see if I can get it working tomorrow.

packages/components/src/Lettera.purs
jllang: I thought using JSON would be consistent with our previous decision to add JSON logging for incoming requests.

packages/components/src/Lettera.purs
jllang: Is `Show` better than `Json` for displaying formatted errors? Or how about using some custom `formatErrorMessage` function so that we wouldn't be using standard classes for a non-standard purpose?

packages/components/src/Lettera.purs
jllang: I improved the type by using `Json` instead of `String`.

packages/components/src/Lettera.purs
jllang: Ok, now we have two `Json` values as well as one `Int`.

packages/components/src/Lettera.purs
jllang: I think that using JSON instead of more human readable ad hoc error messages might make it easier to search for errors in our cloug log system.

packages/components/src/Lettera/Models.purs
```diff
@@ -633,3 +639,105 @@ editorialIdToUuid editorialId =
 url_namespace :: UUID.UUID
 url_namespace = unsafePartial $ fromJust $ UUID.parseUUID "6ba7b811-9dad-11d1-80b4-00c04fd430c8"
 
+encodeRequestHeaders :: Array AX.RequestHeader -> Maybe Json
```
kaol, line 642: I think `Lettera.Models` is too busy already and nothing in here uses these so it's not justified by dependencies either. The focus of this module is diluted if it gets Affjax imports to it. If you're moving these from `Lettera` make it a new module. `Lettera.Requests` or if you have a better idea.

packages/components/src/Lettera/Models.purs
```diff
@@ -633,3 +639,105 @@ editorialIdToUuid editorialId =
...
+      encodeError "HttpError"
+                  (Just letteraErrorRequest)
+                  (Just letteraErrorResponse)
+                  (Just (encodeInt letteraErrorCode))
+    ParseError ->
+      encodeError "ParseErrror" Nothing Nothing Nothing
+
+data LetteraError
+  = ResponseError
+    { letteraErrorRequest  :: Json
```
kaol, line 702: I think I said it already but don't use `Json` on this level. Leave generating it to the `encodeJson` instance. Ditto for `HttpError`. If we need users of this code to do actions based on responses then that's a lot more convenient format.

packages/components/src/Lettera/Models.purs
```diff
@@ -633,3 +639,105 @@ editorialIdToUuid editorialId =
 url_namespace :: UUID.UUID
 url_namespace = unsafePartial $ fromJust $ UUID.parseUUID "6ba7b811-9dad-11d1-80b4-00c04fd430c8"
 
+encodeRequestHeaders :: Array AX.RequestHeader -> Maybe Json
+encodeRequestHeaders [] = Nothing
+encodeRequestHeaders hs =
+  let encodeHeader h obj = flip extend obj $ case h of
+        AX.Accept (MediaType v)      -> "Accept" := v
```
kaol, line 646: If you're doing white spaces for alignment use it on `:=` too.

packages/components/src/Lettera/Models.purs
```diff
@@ -633,3 +639,105 @@ editorialIdToUuid editorialId =
...
+        AX.ContentType (MediaType v) -> "ContentType" := v
+        AX.RequestHeader k v         -> k := v
+  in Just $ foldr encodeHeader jsonEmptyObject hs
```
kaol, line 649: Is this going to include authorization headers in it? I think a better default would be to scrub them from `LetteraResponse` (and if convenient make it easy in the source to enable them for dev use when needed).

kaol (changes requested):

packages/components/src/Lettera/Models.purs
jllang: Okay, but then we'll have more polymorphic type variables to deal with.

jllang: I started to think that maybe including requests and responses in the error type is not the best solution to start with. That information is needed for logging but is it really an essential part of an error value? Perhaps we could consider moving the logging from Mosaico to Affresco to the place where we have the best information about the requests, responses and/or other context?

jllang: It seems that If I generalise the `Json` type occurrences into type variables, then Mosaico and Affresco disagree whether the `getFrontpage` and `getFrontpageHtml` are dealing with `LetteraResponse String String a` or `LetteraResponse Json Json a`.

PR #1489, opened 2023-09-20
Publico font in articles

danielrlc:


apps/app-article-server/README.md
```diff
@@ -8,18 +8,18 @@ Authentication is done via `AuthUser` and `Authorization` headers.
...
-
 ```
 export PUBLIC_URL=/dist
 yarn install
 yarn build
 yarn start
 ```
 
-As an alternative to exporting PUBLIC_URL for every new shell you can also use [direnv](https://direnv.net/docs/installation.html) to have it done automgically.
+As an alternative to exporting PUBLIC_URL for every new shell you can also use [direnv](https://direnv.net/docs/installation.html) to have it done automagically.
```
jllang, line 18: Is there a reason why we can't set it by default?

jllang (approved):

apps/app-article-server/README.md
danielrlc: I don't think so. I've got used to setting it manually myself though.

PR #1496, opened 2023-10-06
Render and style tables from tables plugin in app-article-server

danielrlc:
https://trello.com/c/snRNT62x/1647-styling-for-tables-in-mosaico-and-app-article-server


joakimekroos (changes requested):
We're getting the

`Warning: Each child in a list should have a unique "key" prop.`

When rendering the table elements.

Maybe add an index to the map functions and use that to give them unique "key" props?

danielrlc:
> Maybe add an index to the map functions and use that to give them unique "key" props?

Done. Good you spotted that.

joakimekroos (approved):

PR #1499, opened 2023-10-10
UserCentrics consent

joakimekroos:


apps/app-article-server/src/browser/index.html
```diff
@@ -146,6 +154,8 @@
 
       window.googletag.cmd.push(function () {
         googletag.pubads().setTargeting("sites-or-apps", "apps-only");
+        // googletag.pubads().setPrivacySettings({limitedAds: true});
```
mrskug, line 157: If this is not coming back then it would be better to delete the line instead of commenting out

mrskug (approved):

PR #1500, opened 2023-10-11
Dark mode for tables

danielrlc:


jllang (approved):

apps/app-article-server/src/browser/components/article-content.js
```diff
@@ -263,10 +263,9 @@ class Content extends Component {
 
   renderTable(block, key) {
     return (
-      <div className="responsive-table-wrapper">
+      <div key={key} className="responsive-table-wrapper">
```
jllang, line 266: Was the key on wrong level before?

danielrlc: Yes, I added the wrapper around the table and forgot to move the key up a level so React was complaining.

PR #1505, opened 2023-10-30
Add support for Lettera free passes

jllang:


packages/components/src/Lettera.purs
```diff
@@ -109,19 +109,39 @@ handleLetteraError = Console.warn <<< show
...
-getArticleWithUrl :: String -> Maybe Paper -> Maybe UserAuth -> Maybe String -> Aff (Either String FullArticle)
-getArticleWithUrl url paper auth clientip = do
+getArticleWithUrl
+  :: String
+  -> Maybe Paper
+  -> Maybe UserAuth
+  -> Maybe String
+  -> Maybe String
+  -> Aff (Either String FullArticle)
+getArticleWithUrl url paper auth clientip freePassHash= do
```
kaol, line 140: Add white space.

packages/components/src/Lettera.purs
```diff
@@ -109,19 +109,39 @@ handleLetteraError = Console.warn <<< show
...
   let request = AX.defaultRequest
-        { url = url <> foldMap (\p -> "?paper=" <> Paper.toString p) paper
+        { url = url <>
+                foldMap (\p -> "?paper=" <> Paper.toString p) paper <>
+                foldMap (\f -> "&freePassHash=" <> f) freePassHash
```
kaol, line 144: Is the token guaranteed to only have URL safe characters?

kaol (approved):

packages/components/src/Lettera.purs
jllang: The token comes from URL, but I can add sanitation.

jllang: Is it okay if I sanitize free pass hash already when parsing the URL in Mosaico? I could add a comment that the hash should be sanitized by the caller of  this function.

PR #1517, opened 2023-12-01
Update Persona client

jllang:


kaol (approved):

jllang:
I believe we need either this or https://git.ksfmedia.fi/taco/persona/-/merge_requests/357 for the credentials migration.

